PR #65, opened 2024-04-19
Move voltage level nodes with multiple buses in network area diagram

massimo-ferraro:

no

feature

The voltage levels with multiple buses are not correctly moved

The voltage levels with multiple buses can be moved, with correct computation of edges start positions, arrow position, and a possible redrawing of buses

- [ ] Yes
- [X] No

### DIFF
--- a/src/components/network-area-diagram-viewer/diagram-utils.ts
+++ b/src/components/network-area-diagram-viewer/diagram-utils.ts
@@ -242,3 +242,138 @@ function classIsContainerOfDraggables(element: SVGElement): boolean {
         element.classList.contains('nad-3wt-nodes')
     );
 }
+
+// get inner and outer radius of bus node and radius of voltage level
+export function getNodeRadius(
+    nbNeighbours: number,
+    voltageLevelCircleRadius: number,
+    busIndex: number,
+    interAnnulusSpace: number
+): [number, number, number] {
+    const vlCircleRadius: number =
+        Math.min(Math.max(nbNeighbours + 1, 1), 2) * voltageLevelCircleRadius;
+    const unitaryRadius = vlCircleRadius / (nbNeighbours + 1);
+    return [
+        busIndex * unitaryRadius + interAnnulusSpace / 2,
+        (busIndex + 1) * unitaryRadius - interAnnulusSpace / 2,
+        vlCircleRadius,
+    ];
+}
+
+function getCirclePath(
+    radius: number,
+    angleStart: number,
+    angleEnd: number,
+    clockWise: boolean
+) {
+    const arcAngle = angleEnd - angleStart;
+    const xStart = radius * Math.cos(angleStart);
+    const yStart = radius * Math.sin(angleStart);
+    const xEnd = radius * Math.cos(angleEnd);
+    const yEnd = radius * Math.sin(angleEnd);
+    const largeArc = Math.abs(arcAngle) > Math.PI ? 1 : 0;
+    return (
+        xStart.toFixed(3) +
+        ',' +
+        yStart.toFixed(3) +
+        ' A' +
+        radius.toFixed(3) +
+        ',' +
+        radius.toFixed(3) +
+        ' ' +
+        radToDeg(arcAngle).toFixed(3) +
+        ' ' +
+        largeArc +
+        ' ' +
+        (clockWise ? 1 : 0) +
+        ' ' +
+        xEnd.toFixed(3) +
+        ',' +
+        yEnd.toFixed(3)
+    );
+}
+
+// get path for bus annulus
+export function getFragmentedAnnulusPath(
+    angles: number[],
+    busNodeRadius: [number, number, number],
+    nodeHollowWidth: number
+): string {
+    let path: string = '';
+    const halfWidth = nodeHollowWidth / 2;
+    const deltaAngle0 = halfWidth / busNodeRadius[1];
+    const deltaAngle1 = halfWidth / busNodeRadius[0];
+    for (let index = 0; index < angles.length; index++) {
+        const outerArcStart = angles[index] + deltaAngle0;
+        const outerArcEnd = angles[index + 1] - deltaAngle0;
+        const innerArcStart = angles[index + 1] - deltaAngle1;
+        const innerArcEnd = angles[index] + deltaAngle1;
+        if (outerArcEnd > outerArcStart && innerArcEnd < innerArcStart) {
+            path =
+                path +
+                'M' +
+                getCirclePath(
+                    busNodeRadius[1],
+                    outerArcStart,
+                    outerArcEnd,
+                    true
+                ) +
+                ' L' +
+                getCirclePath(
+                    busNodeRadius[0],
+                    innerArcStart,
+                    innerArcEnd,
+                    false
+                ) +
+                ' Z ';
+        }
+    }
+    return path;
+}
+
+function getPolylinePoints(polylinePoints: string): Point[] | null {
+    const coordinates: string[] = polylinePoints.split(/,| /);
+    if (coordinates.length < 4) {
+        return null;
+    }
+    const points: Point[] = [];
+    for (let index = 0; index < 4; index = index + 2) {
+        const point = new Point(+coordinates[index], +coordinates[index + 1]);
+        points.push(point);
+    }
+    return points;
+}
+
+// get angle of first 2 points of a polyline
+export function getPolylineAngle(polylinePoints: string): number | null {
+    const points: Point[] | null = getPolylinePoints(polylinePoints);
+    if (points == null) {
+        return null;
+    }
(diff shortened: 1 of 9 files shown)


src/components/network-area-diagram-viewer/diagram-utils.ts
```diff
@@ -192,3 +192,138 @@ function classIsContainerOfDraggables(element: SVGElement): boolean {
...
+    for (let index = 0; index < 4; index = index + 2) {
+        const point = new Point(+coordinates[index], +coordinates[index + 1]);
+        points.push(point);
+    }
+    return points;
+}
+
+// get angle of first 2 points of a polyline
+export function getPolylineAngle(polylinePoints: string): number | null {
+    const points: Point[] | null = getPolylinePoints(polylinePoints);
```
flo-dup, line 299: Similarly for polyline
```suggestion
export function getPolylineAngle(polyline: HTMLElement): number | null {
    const pointsString = polyline.getAttribute('points');
    const points: Point[] | null = getPolylinePoints(pointsString);
```

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -733,4 +855,263 @@ export class NetworkAreaDiagramViewer {
...
+                    totalTranslation.y.toFixed(2) +
+                    ')'
+            );
+            // store edge angles, to use them for bus node redrawing
+            this.storeHalfLoopEdgeAngle(edgeNode, edgeNode.id + '.1', 0);
+            this.storeHalfLoopEdgeAngle(edgeNode, edgeNode.id + '.2', 1);
+        });
+    }
+
+    private storeHalfLoopEdgeAngle(
```
flo-dup, line 893: Can't you factorize this method with `storeHalfEdgeAngle`, taking a function `getAngle: (a: HTMLElement) => number` as parameter? which would be either `DiagramUtils::getPathAngle` or `DiagramUtils::getPolylineAngle`.

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -733,4 +855,263 @@ export class NetworkAreaDiagramViewer {
...
+                const busEdges =
+                    busNodeEdges.get(busNode.getAttribute('svgid') ?? '-1') ??
+                    [];
+                busEdges.forEach((edge) => {
+                    const busId = busNode.getAttribute('svgid') ?? '-2';
+                    const busNode1 = edge.getAttribute('busnode1') ?? '-1';
+                    const edgeId = edge.getAttribute('svgid');
+                    const angleId =
+                        busId == busNode1 ? edgeId + '.1' : edgeId + '.2';
+                    const edgeAngle = this.edgeAngles.get(angleId);
```
flo-dup, line 981: I think it would be clearer to call a `getEdgeAngle` method here, instead of computing it before in `redrawOtherVoltageLevelNode` and `moveEdge`. That's the only place where we need those angles.

This `getEdgeAngle` method would either return the value from the `edgeAngles` map or computes it and push it to the map. And if I'm not mistaken we have everything we need to compute them here.

flo-dup (changes requested):

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -733,4 +855,263 @@ export class NetworkAreaDiagramViewer {
...
+            );
+            //transform?.setTranslate(totalTranslation.x, totalTranslation.y);
+            edgeNode?.setAttribute(
+                'transform',
+                'translate(' +
+                    totalTranslation.x.toFixed(2) +
+                    ',' +
+                    totalTranslation.y.toFixed(2) +
+                    ')'
+            );
```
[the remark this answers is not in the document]
massimo-ferraro: method moveSvgElement used here. The moveSvgElement method now returns the moved SVG element, to avoid here to search for it again.

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -733,4 +855,263 @@ export class NetworkAreaDiagramViewer {
...
+                return;
+            }
+            // sort buses by index
+            const sortedBusNodes: SVGGraphicsElement[] = [];
+            busNodes.forEach((busNode) => {
+                const index = busNode.getAttribute('index') ?? '-1';
+                if (+index >= 0) {
+                    sortedBusNodes[+index] = busNode;
+                }
+            });
```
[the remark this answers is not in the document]
massimo-ferraro: function getSortedBusNodes added

src/components/network-area-diagram-viewer/diagram-utils.ts
```diff
@@ -192,3 +192,138 @@ function classIsContainerOfDraggables(element: SVGElement): boolean {
...
+            .split(',');
+        const point = new Point(+coordinates[0], +coordinates[1]);
+        points.push(point);
+    }
+    return points;
+}
+
+// get angle of first 2 points of a path
+export function getPathAngle(pathPoints: string): number | null {
+    const points: Point[] | null = getPathPoints(pathPoints);
```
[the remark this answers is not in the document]
massimo-ferraro: done

src/components/network-area-diagram-viewer/diagram-utils.ts
massimo-ferraro: done

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: getAngle function parameter added to storeHalfEdgeAngle method

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: The angles could be needed to redraw the moved node, and possibly other connected nodes. 
For the moved node, the computation (and storing in the map) of the angles has been put in the moveEdge method because all the data to compute the angle (i.e. the 2 points of the edge polyline) are already available there (the points are computed to move the edge). The position of the code to compute and store the angle for the other edges follow the approach.
I can move all the computation of the angles here, but, at least for the edges connected to the moved node, it would require to retrieve the 2 polyline points used to compute the angle from the dom elements (as it is done when redrawing nodes connected to the moved node), instead of using already available values.
It would improve the readability of the code, loosing something in the performances.

flo-dup: Ok, I understand.

In order not to deteriorate the performances but to improve the readability, I think we could do something in between
- set the values of the angle in advance only if the points are already known, that is, only in `moveEdge` function
- for the other angles, compute them here if `this.edgeAngles.get(angleId)` is undefined in `redrawVoltageLevelNode` function.
 
This will factorize the code calling `storeHalfEdgeAngle`.

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -703,4 +835,241 @@ export class NetworkAreaDiagramViewer {
...
+                edgeNode.id + '.1',
+                0,
+                DiagramUtils.getPathAngle
+            );
+            this.storeHalfEdgeAngle(
+                edgeNode,
+                edgeNode.id + '.2',
+                1,
+                DiagramUtils.getPathAngle
+            );
```
flo-dup, line 872: This can be done later, but I just realize that it's at the cost of requesting again the edgeId. As a compromise readability / performance that should be ok, don't you think?

massimo-ferraro: Moved code used for computing and storing edge angles

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -703,4 +830,211 @@ export class NetworkAreaDiagramViewer {
...
+                edgeNode.children.item(sideIndex)?.firstElementChild
+            );
+            if (halfEdge != null) {
+                const angle = getAngle(halfEdge);
+                if (angle != null) {
+                    this.edgeAngles.set(angleId, angle);
+                }
+            }
+        }
+    }
```
flo-dup, line 869: this is now only called in `getEdgeAngle`, so we could remove this method and include it there directly.

Besides we can remove the `if (!this.edgeAngles.has(angleId))`, as it is already checked before in `getEdgeAngle`

massimo-ferraro: Code moved to getEdgeAngle method

flo-dup (approved):

PR #73, opened 2024-05-30
Update metadata and add callback for node moving in network area diagram

massimo-ferraro:

no

feature

When a node is moved in network area diagram the metadata of the node is not changed, and no callback is available for having information about the node moving

When a node is moved the metadata of the node is updated with the new position, and a callback has been added for having information about the node moving

- [X] Yes
- [ ] No

the constructor of the NetworkAreaDiagramViewer has a new additional parameter: the callback.
An example of callback is provided in the demo/src/diagram-viewers/add-diagrams.js file

### DIFF
--- a/src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
+++ b/src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
@@ -13,6 +13,15 @@ import { SvgParameters } from './svg-parameters';
 type DIMENSIONS = { width: number; height: number; viewbox: VIEWBOX };
 type VIEWBOX = { x: number; y: number; width: number; height: number };
 
+export type OnMoveNodeCallbackType = (
+    equipmentId: string,
+    nodeId: string,
+    x: number,
+    y: number,
+    XOrig: number,
+    yOrig: number
+) => void;
+
 export class NetworkAreaDiagramViewer {
     container: HTMLElement;
     svgContent: string;
@@ -28,6 +37,7 @@ export class NetworkAreaDiagramViewer {
     initialPosition: Point = new Point(0, 0);
     svgParameters: SvgParameters;
     edgeAngles: Map<string, number> = new Map<string, number>();
+    onNodeCallback: OnMoveNodeCallbackType | null;
 
     constructor(
         container: HTMLElement,
@@ -35,7 +45,8 @@ export class NetworkAreaDiagramViewer {
         minWidth: number,
         minHeight: number,
         maxWidth: number,
-        maxHeight: number
+        maxHeight: number,
+        onNodeCallback: OnMoveNodeCallbackType | null
     ) {
         this.container = container;
         this.svgContent = svgContent;
@@ -48,6 +59,7 @@ export class NetworkAreaDiagramViewer {
         // so far, they are hardcoded in the class
         // the idea is to read them from the metadata included in the SVG
         this.svgParameters = new SvgParameters();
+        this.onNodeCallback = onNodeCallback;
     }
 
     public setWidth(width: number): void {
@@ -250,7 +262,7 @@ export class NetworkAreaDiagramViewer {
         if (this.selectedElement) {
             event.preventDefault();
             this.ctm = this.svgDraw?.node.getScreenCTM();
-            this.updateGraph(event);
+            this.updateGraph(event, false);
             this.initialPosition = DiagramUtils.getPosition(
                 this.selectedElement
             );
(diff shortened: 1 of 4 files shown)


src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -1208,4 +1223,30 @@ export class NetworkAreaDiagramViewer {
...
+        // get move node in metadata
+        const node: SVGGraphicsElement | null = this.container.querySelector(
+            'nad\\:node[svgid="' + this.selectedElement?.id + '"]'
+        );
+        if (node != null) {
+            // save original position, for the callback
+            const xOrig = node.getAttribute('x') ?? 0;
+            const yOrig = node.getAttribute('y') ?? 0;
+            // update node position in metadata
+            node.setAttribute('x', mousePosition.x.toFixed(2));
```
flo-dup, line 1237: Here we do this rounding twice instead of introducing a variable. But I realize that the mouse position is rounded to 2 decimals most of the time afterwards (sometimes it's not, but the it's the translation which is rounded). Couldn't we avoid this by simply returning directly this rounded values in `getMousePosition(event)`?

flo-dup (changes requested):

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -1208,4 +1223,30 @@ export class NetworkAreaDiagramViewer {
...
+            node.setAttribute('y', mousePosition.y.toFixed(2));
+            // call the callback, if defined
+            if (this.onNodeCallback != null) {
+                this.onNodeCallback(
+                    node.getAttribute('equipmentid') ?? '',
```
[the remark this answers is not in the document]
massimo-ferraro: The method also calsl the callback because it has all the data for the callback, taken from the metadata. I changed the name

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -288,11 +300,14 @@ export class NetworkAreaDiagramViewer {
...
     }
 
-    private updateGraph(event: Event) {
+    private updateGraph(event: Event, updateMetadata: boolean) {
         const mousePosition = this.getMousePosition(event as MouseEvent);
         this.moveNode(mousePosition);
         this.moveText(mousePosition);
         this.moveEdges(mousePosition);
+        if (updateMetadata) {
+            this.updataMetadata(mousePosition);
```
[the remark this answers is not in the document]
massimo-ferraro: fixed

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -259,7 +271,7 @@ export class NetworkAreaDiagramViewer {
 
     private endDrag(event: Event) {
         if (this.selectedElement) {
-            this.updateGraph(event);
+            this.updateGraph(event, true);
```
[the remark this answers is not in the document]
massimo-ferraro: Refactored

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: Actually the toFixed method does not round the number, it formats the number as string. It is used when setting the data to the svg nodes attributes.  I added a method for formatting these values (and also other 2 methods for points and polyline) to make thing more clear

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -1208,4 +1209,36 @@ export class NetworkAreaDiagramViewer {
...
+            node.setAttribute(
+                'y',
+                DiagramUtils.getFormattedValue(mousePosition.y)
+            );
+            // call the callback, if defined
+            if (this.onNodeCallback != null) {
+                this.onNodeCallback(
+                    node.getAttribute('equipmentid') ?? '',
+                    node.getAttribute('svgid') ?? '',
+                    DiagramUtils.round(mousePosition.x, 2),
```
flo-dup, line 1236: You'd rather reuse the exact same value as in the metadata. Just introduce a variable `const xNew = DiagramUtils.getFormattedValue(mousePosition.x)`, and similarly for y.

flo-dup (changes requested):

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: done

flo-dup (approved):

PR #100, opened 2024-09-24
handle toggle popover when mouseover equipment

souissimai:

handle popover when onhovering equipment

Nothing happen when onhovering equipment from NAD

Show popover when onhovering equipment ( line or twt ) from NAD

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/demo/src/diagram-viewers/add-diagrams.ts
+++ b/demo/src/diagram-viewers/add-diagrams.ts
@@ -318,8 +318,8 @@ const handleBus: OnBusCallbackType = (id, svgId, x, y) => {
     console.log(msg);
 };
 
-const handleToggleSldHover: OnToggleSldHoverCallbackType = (shouldDisplay, anchorEl, equipmentId, equipmentType) => {
-    if (shouldDisplay) {
+const handleToggleSldHover: OnToggleSldHoverCallbackType = (hovered, anchorEl, equipmentId, equipmentType) => {
+    if (hovered) {
         const msg = 'Hovers on equipment: ' + equipmentId + ', equipmentType: ' + equipmentType;
         console.log(msg);
     }
--- a/src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
+++ b/src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
@@ -381,7 +381,6 @@ export class NetworkAreaDiagramViewer {
         }
 
         const hoverableElem = DiagramUtils.getHoverableFrom(mouseEvent.target as SVGElement);
-        //const parentElement = hoverableElem?.parentElement;
         if (!hoverableElem) {
             this.onToggleHoverCallback(false, null, '', '');
             return;
--- a/src/components/single-line-diagram-viewer/single-line-diagram-viewer.ts
+++ b/src/components/single-line-diagram-viewer/single-line-diagram-viewer.ts
@@ -97,7 +97,7 @@ export type OnFeederCallbackType = (
 export type OnBusCallbackType = (busId: string, svgId: string, x: number, y: number) => void;
 
 export type OnToggleSldHoverCallbackType = (
-    shouldDisplay: boolean,
+    hovered: boolean,
     anchorEl: EventTarget | null,
     equipmentId: string,
     equipmentType: string
@@ -120,7 +120,7 @@ export class SingleLineDiagramViewer {
     arrowSvg: string;
     arrowHoverSvg: string;
     svgDraw: Svg | undefined;
-    handleTogglePopover: OnToggleSldHoverCallbackType | null;
+    onToggleHoverCallback: OnToggleSldHoverCallbackType | null;
 
     constructor(
         container: HTMLElement,
@@ -136,7 +136,7 @@ export class SingleLineDiagramViewer {
         onFeederCallback: OnFeederCallbackType | null,
         onBusCallback: OnBusCallbackType | null,
         selectionBackColor: string,
-        handleTogglePopover: OnToggleSldHoverCallbackType | null
+        onToggleHoverCallback: OnToggleSldHoverCallbackType | null
     ) {
         this.container = container;
         this.svgType = svgType;
@@ -155,7 +155,7 @@ export class SingleLineDiagramViewer {
         this.arrowSvg = ARROW_SVG;
         this.arrowHoverSvg = ARROW_HOVER_SVG;
         this.addNavigationArrow();
-        this.handleTogglePopover = handleTogglePopover;
+        this.onToggleHoverCallback = onToggleHoverCallback;
     }
 
     public setWidth(width: number): void {
@@ -586,10 +586,10 @@ export class SingleLineDiagramViewer {
                     equipment.componentType === 'PHASE_SHIFT_TRANSFORMER'
                         ? 'TWO_WINDINGS_TRANSFORMER'
                         : equipment.componentType;
-                this.handleTogglePopover?.(true, event.currentTarget, equipment.equipmentId, equipmentType);
+                this.onToggleHoverCallback?.(true, event.currentTarget, equipment.equipmentId, equipmentType);
             });
             svgEquipment?.addEventListener('mouseout', () => {
-                this.handleTogglePopover?.(false, null, '', '');
+                this.onToggleHoverCallback?.(false, null, '', '');
             });
         });
     }


jamal-khey (commented):
Test Ok

demo/src/diagram-viewers/add-diagrams.ts
```diff
@@ -379,3 +387,23 @@ const handleNodeSelect: OnSelectNodeCallbackType = (equipmentId, nodeId) => {
...
+        const msg =
+            'Hovers on equipment: ' +
+            equipmentId +
+            ', equipmentType: ' +
+            equipmentType +
+            ', mousePosition : x =' +
+            mousePosition?.x +
+            ', y=' +
+            mousePosition?.y;
+        console.log(msg);
```
YenguiSeddik, line 407: delete ?

jamal-khey (approved):
Test and code Ok

YenguiSeddik (approved):
test & code : ok

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -362,6 +374,40 @@ export class NetworkAreaDiagramViewer {
...
+        const hoverableElem = DiagramUtils.getHoverableForm(mouseEvent.target as SVGElement);
+        const parentElement = hoverableElem?.parentElement;
+        if (hoverableElem == null || parentElement == null) {
+            this.handleTogglePopover(false, null, '', '');
+            return;
+        }
+        if (!DiagramUtils.classIsContainerOfHoverables(parentElement)) {
+            this.handleTogglePopover(false, null, '', '');
+            return;
+        }
```
flo-dup, line 392: Couldn't his be simplifified with a single `if (!hoverableElem)`?

Besides, if `!hoverableElem` then it means we're hovering over something of no interest. If I understand properly your need, with the shouldDisplay boolean, and the "toggle" naming, you want to be notified when you just stopped hovering a line for instance. Is that the usecase? I'm not sure to get the sense of shouldDisplay in fact. And I don't see the point to call the callback with no equipment id provided.

demo/src/diagram-viewers/add-diagrams.ts
```diff
@@ -379,3 +387,23 @@ const handleNodeSelect: OnSelectNodeCallbackType = (equipmentId, nodeId) => {
     const msg = 'Node ' + nodeId + ' equipment ' + equipmentId + ' selected';
     console.log(msg);
 };
+
+const HandleToggleNadPopover: HandleToggleNadPopoverType = (
```
flo-dup, line 391: wrong case for a variable: **h**andleToggleNadPopover

flo-dup (changes requested):

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -362,6 +374,40 @@ export class NetworkAreaDiagramViewer {
             }
         }
     }
+    private onHover(mouseEvent: MouseEvent) {
+        const equipmentsWithPopover = ['LINE', 'TWO_WINDINGS_TRANSFORMER', 'PHASE_SHIFT_TRANSFORMER'];
```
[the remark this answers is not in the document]
flo-dup: ok, I understand why you did this, following the same logic as in SLD. Indeed, there's no need to proceed in the case of your project using powsybl, even if I'm surprised you dismiss TieLines / DanglingLines. But anyway you're not the only user of this library. In powsybl we try to remain generic so it can be reused by other application with other similar needs.

src/components/network-area-diagram-viewer/diagram-utils.ts
```diff
@@ -472,3 +475,22 @@ export function getNodeMove(node: SVGGraphicsElement, nodePosition: Point): NODE
     const yNew = getFormattedValue(nodePosition.y);
     return { xOrig: xOrig, yOrig: yOrig, xNew: xNew, yNew: yNew };
 }
+
+function hasParentNode(element: SVGElement): boolean {
```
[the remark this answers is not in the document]
flo-dup: I wasn't speaking of renaming but replacing, it's the opposite which is expected here.

src/components/network-area-diagram-viewer/diagram-utils.ts
```diff
@@ -472,3 +475,22 @@ export function getNodeMove(node: SVGGraphicsElement, nodePosition: Point): NODE
...
+    return hasId(element) && element.parentNode != null;
+}
+
+export function getHoverableForm(element: SVGElement): SVGElement | undefined {
+    if (!hasParentNode(element.parentNode as SVGElement)) {
+        return element;
```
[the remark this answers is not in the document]
flo-dup: with a similar logic:
```typescript
function isHoverable(element: SVGElement): boolean {
    return hasId(element)
        && element.parentNode != null
        && classIsContainerOfHoverables(element.parentNode as SVGElement);
}

export function getHoverableFrom(element: SVGElement): SVGElement | undefined {
    if (isHoverable(element)) {
        return element;
    } else if (element.parentElement) {
        return getHoverableFrom(element.parentNode as SVGElement);
    }
}
```

src/components/network-area-diagram-viewer/diagram-utils.ts
```diff
@@ -472,3 +475,22 @@ export function getNodeMove(node: SVGGraphicsElement, nodePosition: Point): NODE
...
+        return getHoverableForm(element.parentNode as SVGElement);
+    }
+}
+export function getStringEdgeType(edge: SVGGraphicsElement): string | null {
+    const edgeTypeIndex = getEdgeType(edge);
+    if (edgeTypeIndex !== null && edgeTypeIndex in EdgeType) {
+        return EdgeType[edgeTypeIndex];
+    }
+    return null;
+}
```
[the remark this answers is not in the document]
flo-dup: ok, missed the fact that we're in strictNullChecks! We could still do it like that with getEdgeType returning an EdgeType in all cases, adding an UNKNOWN EdgeType. Then this always returns a string.

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
flo-dup: with the logic I provided above for `getHoverableFrom`, hoverableElem is null if it has no parent element.

Ok, you need to know when the hovering stopped, I understand. It would make more sense to give the previous equipmentId and type don't you think? They're probably not needed, but it makes more sense to say what is not anymore hovered.

If you need information, we should probably handle the case of hovering a line and then going directly to another one. You can do that if the lines are crossing. We should probably call twice the callback, first with false and the old id, then with true and the new id?

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -38,6 +38,12 @@ export type OnMoveTextNodeCallbackType = (
 ) => void;
 
 export type OnSelectNodeCallbackType = (equipmentId: string, nodeId: string) => void;
+export type HandleToggleNadPopoverType = (
```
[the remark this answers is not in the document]
souissimai: OnNadToggleHoverCallbackType to rename

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -38,6 +38,12 @@ export type OnMoveTextNodeCallbackType = (
 ) => void;
 
 export type OnSelectNodeCallbackType = (equipmentId: string, nodeId: string) => void;
+export type OnToggleHoverCallbackType = (
+    shouldHover: boolean,
```
flo-dup, line 42:
```suggestion
    hovered: boolean,
```

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -63,6 +69,7 @@ export class NetworkAreaDiagramViewer {
     onSelectNodeCallback: OnSelectNodeCallbackType | null;
     shiftKeyOnMouseDown: boolean = false;
     dynamicCssRules: CSS_RULE[];
+    OnToggleHoverCallback: OnToggleHoverCallbackType | null;
```
flo-dup, line 72:
```suggestion
    onToggleHoverCallback: OnToggleHoverCallbackType | null;
```

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -38,6 +38,12 @@ export type OnMoveTextNodeCallbackType = (
 ) => void;
 
 export type OnSelectNodeCallbackType = (equipmentId: string, nodeId: string) => void;
+export type OnToggleHoverCallbackType = (
```
flo-dup, line 41: You noted in a previous comment that you would rename it on SingleLineDiagramViewer side also. Could you do it?
Hence, to avoid clashing with the similar callback from the other viewer, we should add "Nad" keyword inside the type (in the type and in the demo only, the variable names inside the viewers shouldn't clash)

flo-dup (changes requested):

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
souissimai: done

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
souissimai: done

flo-dup (approved):

src/components/single-line-diagram-viewer/single-line-diagram-viewer.ts
```diff
@@ -120,7 +120,7 @@ export class SingleLineDiagramViewer {
     arrowSvg: string;
     arrowHoverSvg: string;
     svgDraw: Svg | undefined;
-    handleTogglePopover: HandleTogglePopoverType | null;
+    handleTogglePopover: OnToggleSLDHoverCallbackType | null;
```
flo-dup, line 123: you'd better change the variable name also. And acronym should not have full upper case in variable/class names
```suggestion
    onToggleHoverCallback: OnToggleSldHoverCallbackType | null;
```

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -38,6 +38,12 @@ export type OnMoveTextNodeCallbackType = (
 ) => void;
 
 export type OnSelectNodeCallbackType = (equipmentId: string, nodeId: string) => void;
+export type OnToggleNADHoverCallbackType = (
```
flo-dup, line 41: Acronym should not have full upper case in variable/class 
```suggestion
export type OnToggleNadHoverCallbackType = (
```

flo-dup (changes requested):
Just two minor comments left!

src/components/single-line-diagram-viewer/single-line-diagram-viewer.ts
flo-dup: You forgot to change the variable name, I'll take care of it.

flo-dup (approved):

PR #106, opened 2024-10-07
[Typescript] Fix OnBreakerCallbackType type

jamal-khey:

FIX typescript type

No Behavior change

No Behavior change

- [x] Yes
- [ ] No

- [x] The *Breaking Change* or *Deprecated* label has been added
- [x] The migration steps are described in the following section

When instantiating the `SingleLineDiagramViewer` class, you need to update the parameter type of the `onBreakerCallback`. The type of the switch element parameter is now `switchElement` instead of the previous type.

**Other information**:


src/components/single-line-diagram-viewer/single-line-diagram-viewer.ts
```diff
@@ -485,7 +485,7 @@ export class SingleLineDiagramViewer {
                         }
                         const switchId = aSwitch.equipmentId;
                         const open = aSwitch.open;
-                        this.onBreakerCallback?.(switchId, !open, event.currentTarget);
```
flo-dup, line 488: isn't the definition of `SvgInHtml` missing? The error can be seen in the CI (when building), but it doesn't fail I don't know why.

flo-dup (changes requested):

src/components/single-line-diagram-viewer/single-line-diagram-viewer.ts
jamal-khey: true i don't know why XD
the type was not correct, i changed it to SVGElement

PR #109, opened 2024-10-30
Fix line overload color computation

jamal-khey:

This PR fixes the function that computes the color of lines based on their overload status. 

Bug fix: Corrects line coloring calculation by considering absolute values of line intensities

The current implementation uses raw intensity values to determine line loading zones, which causes incorrect coloring when the intensity is negative. Negative values are being treated as overload  when they should be treated based on their magnitude regardless of direction.

The fix uses the absolute value of the intensity (Math.abs(intensity)) when comparing against thresholds. This ensures that line loading zones are correctly calculated regardless of the current flow direction. For example, an intensity of -900A on a line with a 1000A limit will now properly show as being close to overload, whereas before it might have been incorrectly shown as safe.

- [ ] Yes
- [x] No

**Other information**:

This fix ensures consistent line coloring behavior regardless of power flow direction
The change affects the visual representation only and does not impact any underlying calculations or data


src/components/network-map-viewer/network/line-layer.js
```diff
@@ -61,9 +61,10 @@ export function getLineLoadingZoneOfSide(limit, intensity, lineFlowAlertThreshol
         return LineLoadingZone.UNKNOWN;
     } else {
         let threshold = (lineFlowAlertThreshold * limit) / 100;
-        if (intensity > 0 && intensity < threshold) {
+        const absoluteIntensity = Math.abs(intensity);
+        if (absoluteIntensity > 0 && absoluteIntensity < threshold) {
```
antoinebhs, line 65: absoluteIntensity is always >0

antoinebhs, line 65: absolute intensity is always > 0 as we checked if intensity === 0 before

YenguiSeddik (approved):

souissimai (approved):

PR #112, opened 2024-11-05
Replace release workflow and add node token as input

antoinebhs:

No

In the publish workflow, it was necessary to generate a release notes to publish on npmjs, now running the action generate the release note and publish on npmjs. See README for details.
It also avoid using credentials of one person and node auth token is added as input (and masked for the rest of the action as per https://dev.to/leading-edje/masking-input-parameters-in-github-actions-1ci) waiting for a native implementation of this in github actions (https://github.com/orgs/community/discussions/12764)

This is based on the work of @TheMaskedTurtle in commons-ui (https://github.com/gridsuite/commons-ui/blob/main/.github/workflows/release.yml)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,7 +29,6 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@v4
         with:
-          fetch-depth: 0
           token: ${{ steps.app-token.outputs.token }}
 
       - uses: actions/setup-node@v4


antoinebhs:
It it works, we can probably also share the workflows of commons-ui and this repo in https://github.com/powsybl/github-ci/tree/main/.github/workflows

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+        id: app-token
+        name: Generate app token
+        with:
+          app-id: ${{ vars.POWSYBL_ACTIONS_APPID }}
+          private-key: ${{ secrets.POWSYBL_ACTIONS_SECRET }}
+
+      - name: Checkout sources
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
flo-dup, line 32: Do we need all branches and tags? 
> Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0 to fetch all history for all branches and tags.

```suggestion
```

flo-dup (approved):

PR #115, opened 2024-11-05
Disable node interaction when no metadata is available in network area diagram viewer

massimo-ferraro:

no

Bug fix

In network area diagram viewer, the node interaction can be enabled even if there is no metadata available

In network area diagram viewer, when there is no metadata available the node interaction is disabled

- [ ] Yes
- [X] No

### DIFF
--- a/src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
+++ b/src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
@@ -97,9 +97,15 @@ export class NetworkAreaDiagramViewer {
         this.originalWidth = 0;
         this.originalHeight = 0;
         this.dynamicCssRules = customDynamicCssRules ?? DEFAULT_DYNAMIC_CSS_RULES;
-        // if no metadata is available disable node interaction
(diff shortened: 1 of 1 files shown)


src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -97,6 +97,8 @@ export class NetworkAreaDiagramViewer {
         this.originalWidth = 0;
         this.originalHeight = 0;
         this.dynamicCssRules = customDynamicCssRules ?? DEFAULT_DYNAMIC_CSS_RULES;
+        // if no metadata is available disable node interaction
+        enableNodeInteraction = diagramMetadata == null ? false : enableNodeInteraction;
```
flo-dup, line 101: We should also disable the `svgDraw.on('mouseover', ...)`. For that, shouldn't we rather give a `hasMetadata` boolean to `init` method?

massimo-ferraro: Hover callback disabled if there is no metadata

flo-dup (approved):

PR #119, opened 2024-11-05
locate substation with zoom

jamal-khey:

centering the view on a substation did not trigger any zoom action on the substation. The camera may have remained at a default zoom level , making it difficult for users to focus on the specific substation they were trying to locate i a dense network

now when the location locating a substation, the camera automatically zooms in on the substation. This provides immediate visual feedback and helps users quickly identify the substation's location on the map.

- [ ] Yes
- [x] No

a new optional props is available `locateSubStationZoomLevel` wich allow to define zoom level.
if not defined this props has a default value to 12


souissimai:
Maybe we need to check if the current zoom level is greater than the default zoom level defined in the code, and then consider the current zoom level. Otherwise, we should use the default one.

jamal-khey:
> Maybe we need to check if the current zoom level is greater than the default zoom level defined in the code, and then consider the current zoom level. Otherwise, we should use the default one.

Done , i keep the max between previous and new zoom level

souissimai (approved):
tested, code OK

YenguiSeddik (approved):

PR #120, opened 2024-11-05
Fix workflow permissions

antoinebhs:

No

Fix permission in workflow after failing to publish release notes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,9 @@ on:
         required: true
         type: string
 
+permissions:
+  contents: write
+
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -16,6 +16,9 @@ on:
         required: true
         type: string
 
+permissions:
```
flo-dup, line 19: I think you can give restrict this permission to the last step only.

antoinebhs: Ok I thought permissions were action scoped, I will check

PR #140, opened 2024-12-13
Don't update dynamiccssrules on translations, add chromium workaround

jonenst:

NO

feature/bugfix

css rules are updated when the viewbox changes, even for translations. But we only use the zoom level in rules
Also, chromium doesn't update the rendering when of percents (stroke-width), so add a workaround to force redraw
when the zoom changes by 1 percent

no update of rules during translations. Chromium redraw on zoom workaround (enabled all the time for consistency)

- [ ] Yes
- [x] No

**Other information**:

we could split this PR in 2. We could enable the workaround only if we detect dynamic rules with percentages


src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -1372,6 +1394,15 @@ export class NetworkAreaDiagramViewer {
 
     public checkAndUpdateLevelOfDetail(svg: SVGSVGElement) {
         const maxDisplayedSize = this.getCurrentlyMaxDisplayedSize();
+        const previousMaxDisplayedSize = this.getPreviousMaxDisplayedSize();
+        // in case of bad or unset values NaN or Infinity, this condition is skipped and the function behaves as if zoom changed
+        if (
+            Math.abs(previousMaxDisplayedSize - maxDisplayedSize) / previousMaxDisplayedSize <
```
EstherDarkish, line 1400: divide by zero

jonenst: in case of bad or unset values NaN or Infinity, this condition is skipped and the function behaves as if zoom changed

YenguiSeddik (approved):

EstherDarkish (approved):

PR #143, opened 2025-01-02
Refactor move SVG Elements

sBouzols:
create a moveElement private function, refactor onDragEnd and remove boolean parameters for callbacks.

We always need to call callbacks on onDragEnd not when moving elements programmatically.


EstherDarkish (approved):
Test OK

souissimai (approved):
Already tested

PR #145, opened 2025-01-10
Handle InsertNameDesc SVG parameter in NAD viewer

massimo-ferraro:

no

Bug fix

Th Network Area Diagram Viewer does not correctly handle nodes moving when the SVG contains node name descriptions (the InsertNameDesc SVG parameter is set to true)

Th Network Area Diagram Viewer correctly handles nodes moving when the SVG contains node name descriptions (the InsertNameDesc SVG parameter is set to true)

- [ ] Yes
- [X] No


src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -1169,7 +1170,8 @@ export class NetworkAreaDiagramViewer {
             busNodeRadius,
             this.svgParameters.getNodeHollowWidth()
         );
-        const busElement: HTMLElement | null = <HTMLElement>node.children.item(busIndex);
+        const childIndex = this.svgParameters.getInsertNameDesc() ? busIndex + 1 : busIndex;
+        const busElement: HTMLElement | null = <HTMLElement>node.children.item(childIndex);
```
flo-dup, line 1174: We can instead do something like:
```typescript
node.querySelector('.nad-busnode:nth-of-type(' + busIndex + ')')`;
```

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -1131,7 +1131,8 @@ export class NetworkAreaDiagramViewer {
             const edgeNode: SVGGraphicsElement | null = this.container.querySelector("[id='" + edgeId + "']");
             if (edgeNode) {
                 const side = busNode.svgId == edge.busNode1 ? 0 : 1;
-                const halfEdge: HTMLElement = <HTMLElement>edgeNode.children.item(side)?.firstElementChild;
+                const childIndex = this.svgParameters.getInsertNameDesc() ? side + 1 : side;
```
flo-dup, line 1134: You could use the half-edge id (called angleId here, maybe we should rename it!) to query directly the half edge element.

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: Used query selector instead of children list, but not nth-of-type, it seems to work with elements of the same types, while here we have 'circle' and 'path' elements

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: Used query selector instead of children list, and renamed variables

src/components/network-area-diagram-viewer/svg-parameters.ts
```diff
@@ -111,4 +112,8 @@ export class SvgParameters {
     public getCurrentValuePrecision(): number {
         return this.svgParametersMetadata?.currentValuePrecision ?? SvgParameters.CURRENT_VALUE_PRECISION_DEFAULT;
     }
+
+    public getInsertNameDesc(): boolean {
```
flo-dup, line 116: not anymore used

PR #146, opened 2025-01-14
demo: restore tests of enableLevelOfDetail

jonenst:

NO

bugfix, restore demo tests

no test for enablelevelofdetail

tests for enablelevelofdetail restored in demo

- [ ] Yes
- [x] No

**Other information**:

spurious change from e79a67fe42

### DIFF
--- a/demo/src/diagram-viewers/add-diagrams.ts
+++ b/demo/src/diagram-viewers/add-diagrams.ts
@@ -268,7 +268,7 @@ export const addNadToDemo = () => {
                 handleTextNodeMove,
                 handleNodeSelect,
                 true,
-                false,
+                true,
                 null,
                 handleToggleNadHover
             );


flo-dup (commented):
Sorry I missed that! But wasn't the enableLevelOfDetail test inserted by mistake in this test? Indeed, this 6-node diagram with only 2 nominal voltages is not appropriate to test it.

massimo-ferraro (changes requested):
Sorry, my fault for not explaining the change I did in the other PR. The problem is that setting this parameter value  to true makes the edge info (e.g. the active power of the lines) disappear in all the demos (e.g. the first example, the Eurostag network, should have the active power shown). It happens only if you set this parameter to true in the last demo example, could it be set in another demo example?

flo-dup (approved):

massimo-ferraro (approved):

PR #153, opened 2025-01-20
Fixes a positioning issue when dragging a textnode

EstherDarkish:

No

Bug fix

The public function that asks to move a textnode to a position did not take into account the textnode's size. Because of this, the textnode was moved to its top left position instead of its center (where the mouse is as the user drags and drops the textnode).

When using the public moveTextNodeToCoordinates function to move a textnode to a given position, the text node is moved using its center instead of its top left position. This makes the movement behave as if the user drags and drops the textnode.

- [ ] Yes
- [x] No


massimo-ferraro (approved):

sBouzols (approved):
Code Review OK
Tests OK
Console warning check OK

PR #174, opened 2025-03-11
move deck.gl layers in their own package with peerDependencies only o…

amarant:
network-map-layers with peerDependencies only on deck.gl and luma.gl for simpler consumption

minor changes :
add check command to validate typescript
add arethetypeswrong cli to check if the typescript export is correct
fix ts export accordingly

No

move the deck.gl layers to a workspace package network-map-layers that only depends on minimal deck.gl peer dependencies

currently a user has to import the main package and deal with issues around potentially conflicting maplibre, react or mui dependencies and declare overrides

it is supposed to keep intact the current usage of the main package

- [ ] Yes
- [x] No


vite.config.ts
```diff
@@ -24,6 +25,11 @@ export default defineConfig((config) => ({
         dts({
             include: ['src'],
             exclude: ['**/*.{spec,test}.{ts,tsx}'],
+            rollupTypes: true,
+            tsconfigPath: './tsconfig.json',
+            afterBuild: () => {
+                copyFileSync('dist/index.d.ts', 'dist/index.d.cts');
```
Tristan-WorkGH, line 31: The .d.ts use `export` keyword which is ESM and not CommonJS `module.exports` pattern. Thsi is not a 

After reading <https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html>, we just need to refer to the same dts for both ESM and CJS files.
```json
      "require": {
        "types": "./dist/index.d.ts",
        "default": "./dist/powsybl-network-viewer.umd.cjs"
      }
```

amarant: It is advised to not use the same file for ESM and CJS typescript definitions, even if they are identical :

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#:~:text=It%E2%80%99s%20important%20to%20note,users%20of%20the%20package.

this is also why I added the attw utility :
https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseESM.md#common-causes

Tristan-WorkGH: Ohh! I didn't known that! 🤔

Well okay for the double declaration part, but copying the ESM `.d.ts`still wouldn't work for CJS `.d.cts` file no?

amarant: Typescript doesn't care if the d.ts declaration are in ESM for a CJS JS, for example in the unit test of tsup (a widely used bundler for libs) [they snapshot identical d.ts and d.mts](https://github.com/egoist/tsup/blob/main/test/__snapshots__/dts.test.ts.snap#L255-L272).

.gitignore
```diff
@@ -1,9 +1,11 @@
 # dependencies
 /node_modules
+/packages/network-map-layers/node_modules
 
 # production
 /dist
 /powsybl-network-viewer-*.tgz
+/packages/network-map-layers/dist
```
sBouzols, line 8:
```suggestion
**/dist
/powsybl-network-viewer-*.tgz
```

package.json
```diff
@@ -41,13 +49,17 @@
...
+        "packages/network-map-layers"
+    ],
     "overrides": {
         "@deck.gl/mapbox": {
             "@types/mapbox-gl": "npm:mapbox-gl@^3.9.4"
         }
     },
     "dependencies": {
         "@mapbox/mapbox-gl-draw": "<1.5.0",
+        "@powsybl/network-map-layers": "workspace:*",
```
sBouzols, line 62: seems strange this `workspace` prefix ?

CBiasuzzi:
Thanks  in advance for your help in understanding how can I use this PR: I'm not familiar with the 'workspace' feature.

I guess that, eventually,  there will be two artifacts (@powsybl/network-map-layers and @powsybl/network-viewer)  published to the npm registry; Is it correct?

I'm trying to build  the two libraries on my local machine (with node v22.14.0, npm 10.9.2), and use them in my project (my project integrates all the three viewers):

```
npm install
npm run build:all
npm pack -w @powsybl/network-map-layers
npm pack
cp .tgz /tmp
```

Note that it builds Ok, and the integrated network-viewer demo works as expected (after a 'npm start').

However,  when I run a 'npm install' in my project's directory, where I've previously edited the package.json by replacing the existing dependency with:

```
"dependencies": {
 "@powsybl/network-map-layers": "file:/tmp/powsybl-network-viewer-1.12.0-dev.0.tgz",
 "@powsybl/network-viewer": "file:/tmp/powsybl-network-viewer-1.12.0-dev.0.tgz"
},
```


I get this error

```
npm error code EUNSUPPORTEDPROTOCOL
npm error Unsupported URL Type "workspace:": workspace:
```


is there a more correct way to use the new dependencies, outside the powsybl-network-viewer project structure and its  demo (where it seems to work great) ?

PR #179, opened 2025-03-18
Keep CSS rules static

### LINKED ISSUE #154: Keep CSS rules static

Currently, following #78, CSS is modified dynamically, when user is zooming in / out for performance reasons. This showed great performance improvement but it appears to be problematic:
- it injects CSS rules in the middle of the page, which lead to a bleeding problem: #149 (see the example in the demo #148).
- it introduced the concept of `CSS_RULE`, which the user has to define in the constructor, which intersects the CSS rules concept.
- it broke the usual CSS rules definition, which usually solely focusing on styling (and not on behavior)
- it seems to bring maintainability issues? several fixes were already needed on this topic (#147, #140)

The CSS remains static, when user is zooming in / out, a zoom level class is added depending on thresholds given by the user as input. The class could be named based on the min/max threshold (e.g. `zoom-level-300-500`), or on the interval index (e.g. `zoom-level-2`).

### Describe the motivation

Maintainability, fixing the #149 issue

### Extra Information

_No response_

quinarygio:

Fixes #154

- [ ] Yes
- [ ] No

Users should define the zoom level CSS classes in the application's stylesheet.
An example stylesheet that replicates the behavior of the previous dynamic CSS rules using class definitions can be found in the demo/src/zoom.css file.

**Other information**:

### DIFF
--- a/src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
+++ b/src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
@@ -1828,7 +1828,6 @@ export class NetworkAreaDiagramViewer {
         if (!elementData) {
             return;
         }
-        this.ctm = this.svgDraw?.node.getScreenCTM();
         const mousePosition: Point = this.getMousePosition(event);
         this.onRightClickCallback?.(elementData.svgId, elementData.equipmentId, elementData.type, mousePosition);
     }
--- a/.gitignore
+++ b/.gitignore
@@ -1,12 +1,9 @@
 # dependencies
 /node_modules
 
-# distribution
+# production
 /dist
 /powsybl-network-viewer-*.tgz
-
-# distrib for demo
-/output
 
 # webstorm
 /.idea
@@ -14,14 +11,8 @@
 # vscode
 /.vscode
 
-# parcel
-/.parcel-cache
+# vite
+/vite.config.ts-timestamp-*
 
-# demo
-/demo/node_modules
-/demo/build
-/demo/dist
-
-# lcov reporst
+# jest
 /coverage
-
--- a/demo/src/App.tsx
+++ b/demo/src/App.tsx
@@ -35,20 +35,6 @@ export default function App() {
         palette: {
             mode: 'dark',
         },
-        // @ts-expect-error: TODO not present in ThemeOptions
-        link: {
-            color: 'green',
-        },
-        node: {
-            background: '#1976d2',
-            hover: '#90caf9',
-            border: '#cce3f9',
-        },
-        selectedRow: {
-            background: '#545C5B',
-        },
-        mapboxStyle: 'mapbox://styles/mapbox/dark-v9',
-        aggrid: 'ag-theme-alpine-dark',
     });
 
     //declare data to be displayed: coordinates and network data
--- a/src/components/network-map-viewer/network/network-map.tsx
+++ b/src/components/network-map-viewer/network/network-map.tsx
@@ -89,8 +89,9 @@ const PICKING_RADIUS = 5;
 
 const CARTO = 'carto';
 const CARTO_NOLABEL = 'cartonolabel';
+const ETALAB = 'etalab';
 const MAPBOX = 'mapbox';
-type MapLibrary = typeof CARTO | typeof CARTO_NOLABEL | typeof MAPBOX;
+type MapLibrary = typeof CARTO | typeof CARTO_NOLABEL | typeof ETALAB | typeof MAPBOX;
 
 const LIGHT = 'light';
 const DARK = 'dark';
@@ -663,6 +664,12 @@ const NetworkMap = forwardRef<NetworkMapRef, NetworkMapProps>((rawProps, ref) =>
                     return 'https://basemaps.cartocdn.com/gl/positron-nolabels-gl-style/style.json';
                 } else {
                     return 'https://basemaps.cartocdn.com/gl/dark-matter-nolabels-gl-style/style.json';
+                }
+            case ETALAB:
+                if (mapTheme === LIGHT) {
+                    return 'https://openmaptiles.geo.data.gouv.fr/styles/positron/style.json';
+                } else {
+                    return 'https://openmaptiles.geo.data.gouv.fr/styles/dark-matter/style.json';
                 }
             default:
                 return 'https://basemaps.cartocdn.com/gl/positron-gl-style/style.json';


demo/src/style.css
```diff
@@ -116,7 +116,291 @@ path.nad-arrow-in:not(.nad-state-in .nad-arrow-in) {visibility: hidden}
 .nad-branch-edges .nad-overload .nad-edge-path {animation: line-blink 3s infinite}
 .nad-vl-nodes .nad-overvoltage {animation: node-over-blink 3s infinite}
 .nad-vl-nodes .nad-undervoltage {animation: node-under-blink 3s infinite}
+.nad-zoom-0 {
 
+    .nad-branch-edges .nad-edge-path,
```
flo-dup, line 121: This is not standard CSS syntax, you need to use pseudo-selector `:is(` instead.

demo/src/style.css
```diff
@@ -116,7 +116,291 @@ path.nad-arrow-in:not(.nad-state-in .nad-arrow-in) {visibility: hidden}
...
+    .nad-3wt-edges .nad-edge-path,
+    .nad-branch-edges .nad-winding,
+    .nad-3wt-nodes .nad-winding,
+    .nad-vl-nodes circle.nad-unknown-busnode {
+        stroke-width: 3
+    }
+    .nad-edge-infos,
+    .nad-label-box,
+    .nad-text-edges,
+    [class^="nad-vl0to30"],
```
flo-dup, line 131: This factorization is probably not improving performance, nor readability. For now you need to write something very verbose like below. Which probably is pushing towards a new design in powsybl-diagram, wih a unique class to give the base voltage range, and the defining ring colors with another class, leading to `.nad-vl300to500 .nad-ring-6 {--nad-vl-color: #f44336}` for instance.
```css
.nad-zoom-0 :is(
    .nad-edge-infos, .nad-label-box, .nad-text-edges,
    .nad-vl0to30-0,.nad-vl0to30-1,.nad-vl0to30-2,.nad-vl0to30-3,.nad-vl0to30-4,.nad-vl0to30-5,.nad-vl0to30-6,.nad-vl0to30-7,.nad-vl0to30-8,.nad-vl0to30-line,
    .nad-vl30to50-0,.nad-vl30to50-1,.nad-vl30to50-2,.nad-vl30to50-3,.nad-vl30to50-4,.nad-vl30to50-5,.nad-vl30to50-6,.nad-vl30to50-7,.nad-vl30to50-8,.nad-vl30to50-line,
    .nad-vl50to70-0,.nad-vl50to70-1,.nad-vl50to70-2,.nad-vl50to70-3,.nad-vl50to70-4,.nad-vl50to70-5,.nad-vl50to70-6,.nad-vl50to70-7,.nad-vl50to70-8,.nad-vl50to70-line,
    .nad-vl70to120-0,.nad-vl70to120-1,.nad-vl70to120-2,.nad-vl70to120-3,.nad-vl70to120-4,.nad-vl70to120-5,.nad-vl70to120-6,.nad-vl70to120-7,.nad-vl70to120-8,.nad-vl70to120-line,
    .nad-vl120to180-0,.nad-vl120to180-1,.nad-vl120to180-2,.nad-vl120to180-3,.nad-vl120to180-4,.nad-vl120to180-5,.nad-vl120to180-6,.nad-vl120to180-7,.nad-vl120to180-8,.nad-vl120to180-line,
    .nad-vl180to300-0,.nad-vl180to300-1,.nad-vl180to300-2,.nad-vl180to300-3,.nad-vl180to300-4,.nad-vl180to300-5,.nad-vl180to300-6,.nad-vl180to300-7,.nad-vl180to300-8,.nad-vl180to300-line
    ) {
    display: block
}
```

demo/src/style.css
```diff
@@ -116,7 +116,291 @@ path.nad-arrow-in:not(.nad-state-in .nad-arrow-in) {visibility: hidden}
...
+    [class*=" nad-vl70to120"],
+    [class^="nad-vl120to180"],
+    [class*=" nad-vl120to180"],
+    [class^="nad-vl180to300"],
+    [class*=" nad-vl180to300"] {
+        display: block
+    }
+    .nad-disconnected .nad-edge-path {
+        stroke-dasharray: 0.5%, 0.5%
+    }
```
flo-dup, line 240: I don't see the point to define this in percentage, but maybe I'm missing something? Could you remove anything expressed in percentage? The svg has been drawn in powsybl-diagram for some values which might be changed a bit, but changing that much the dimensions is spoiling the this makes the diagram quite unattractive.

src/components/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -1613,6 +1503,28 @@ export class NetworkAreaDiagramViewer {
         }
     }
 
+    private getZoomLevel(maxDisplayedSize: number): number {
+        if (maxDisplayedSize >= 20000) {
```
flo-dup, line 1507: Could you use an array instead of these separated values?
Wouldn't it make sense to have that array as input (with a default hardcoded array)? instead of an hardcoded array.

flo-dup (changes requested):

demo/src/style.css
quinarygio: I guess percentage was used to allow for smooth transition without defining many zoom levels css classes only to change line size. We can discuss this.

flo-dup: Ok, we could put this in a separate PR, so that we first keep the same behaviour with this change

flo-dup (approved):

flo-dup (changes requested):
Could you add some doc lines to the constructor to explain the added array parameter?

flo-dup:
> Why are the css rules that was in `dynamic-css-utils.ts` are moved in the demo? Projects that was using this feature will need to copy all theses rules in there project.

You mean to provide a default css file which corresponds to the previous rules and that could be easily imported by the user? Then it would only makes sense for the default array. But yes I agree we should add such a file, it's still handy to have a ready-to-use feature.

flo-dup (approved):

demo/src/style.css
```diff
@@ -339,6 +339,234 @@ path.nad-arrow-in:not(.nad-state-in .nad-arrow-in) {
...
+    .nad-edge-infos, .nad-label-box, .nad-text-edges,
+    .nad-vl0to30-0, .nad-vl0to30-1, .nad-vl0to30-2, .nad-vl0to30-3, .nad-vl0to30-4, .nad-vl0to30-5, .nad-vl0to30-6, .nad-vl0to30-7, .nad-vl0to30-8, .nad-vl0to30-line,
+    .nad-vl30to50-0, .nad-vl30to50-1, .nad-vl30to50-2, .nad-vl30to50-3, .nad-vl30to50-4, .nad-vl30to50-5, .nad-vl30to50-6, .nad-vl30to50-7, .nad-vl30to50-8, .nad-vl30to50-line,
+    .nad-vl50to70-0, .nad-vl50to70-1, .nad-vl50to70-2, .nad-vl50to70-3, .nad-vl50to70-4, .nad-vl50to70-5, .nad-vl50to70-6, .nad-vl50to70-7, .nad-vl50to70-8, .nad-vl50to70-line,
+    .nad-vl70to120-0, .nad-vl70to120-1, .nad-vl70to120-2, .nad-vl70to120-3, .nad-vl70to120-4, .nad-vl70to120-5, .nad-vl70to120-6, .nad-vl70to120-7, .nad-vl70to120-8, .nad-vl70to120-line,
+    .nad-vl120to180-0, .nad-vl120to180-1, .nad-vl120to180-2, .nad-vl120to180-3, .nad-vl120to180-4, .nad-vl120to180-5, .nad-vl120to180-6, .nad-vl120to180-7, .nad-vl120to180-8, .nad-vl120to180-line,
+    .nad-vl180to300-0, .nad-vl180to300-1, .nad-vl180to300-2, .nad-vl180to300-3, .nad-vl180to300-4, .nad-vl180to300-5, .nad-vl180to300-6, .nad-vl180to300-7, .nad-vl180to300-8, .nad-vl180to300-line
+    ) {
+    display: block
```
flo-dup, line 363: Can't you remove all the `display: block` rules?

demo/src/style.css
```diff
@@ -339,6 +339,234 @@ path.nad-arrow-in:not(.nad-state-in .nad-arrow-in) {
...
+    stroke-dasharray: 10 10
+}
+.nad-zoom-1000 :is(
+    .nad-branch-edges .nad-edge-path,
+    .nad-3wt-edges .nad-edge-path,
+    .nad-branch-edges .nad-winding,
+    .nad-3wt-nodes .nad-winding,
+    .nad-vl-nodes circle.nad-unknown-busnode) {
+    stroke-width: 0.25%
+}
```
flo-dup, line 376: this rules are common for most of the zooms, so you could have it as default, in order to see better what's changing

demo/src/style.css
```diff
@@ -339,6 +339,234 @@ path.nad-arrow-in:not(.nad-state-in .nad-arrow-in) {
...
+    .nad-vl70to120-0, .nad-vl70to120-1, .nad-vl70to120-2, .nad-vl70to120-3, .nad-vl70to120-4, .nad-vl70to120-5, .nad-vl70to120-6, .nad-vl70to120-7, .nad-vl70to120-8, .nad-vl70to120-line,
+    .nad-vl120to180-0, .nad-vl120to180-1, .nad-vl120to180-2, .nad-vl120to180-3, .nad-vl120to180-4, .nad-vl120to180-5, .nad-vl120to180-6, .nad-vl120to180-7, .nad-vl120to180-8, .nad-vl120to180-line,
+    .nad-vl180to300-0, .nad-vl180to300-1, .nad-vl180to300-2, .nad-vl180to300-3, .nad-vl180to300-4, .nad-vl180to300-5, .nad-vl180to300-6, .nad-vl180to300-7, .nad-vl180to300-8, .nad-vl180to300-line
+    ) {
+    display: block
+}
+
+.nad-zoom-2500 .nad-disconnected .nad-edge-path {
+    stroke-dasharray: 0.5% 0.5%
+}
```
flo-dup, line 417: Same for this stroke

flo-dup (changes requested):

src/components/network-area-diagram-viewer/dynamic-css-utils.ts
Tristan-WorkGH, line 1: Please keep the rules available to projects currently using them, not just in the demo.

Tristan-WorkGH (changes requested):

src/components/network-area-diagram-viewer/dynamic-css-utils.ts
Tristan-WorkGH: As @flo-dup suggest, either as default that can be override or at least as importable rules or css.

flo-dup: About the first suggestion, in fact I don't think we should put any default CSS in the component. That would mean injecting some CSS? Then how would a webpage with two components work, there would be conflicts with the two CSS?

Instead as you said we could give one or several CSS examples. But to me it's not clear how we could include a CSS file in the release.

Another simpler way to do this would be to leave the examples in the demo, and just add in the next migration guide the CSS which corresponds to the previous default `DEFAULT_DYNAMIC_CSS_RULES: CSS_RULE[]`. Indeed it's just a matter of migration, isn't it?

EstherDarkish (approved):

Tristan-WorkGH (approved):
Let's put in the migration guide a comment for the changes and I'm good @flo-dup.

Also tested for GridSuite by @EstherDarkish and good to go.

flo-dup (approved):

PR #223, opened 2025-09-09
[nad-viewer] Dismiss injections when drawing rings

flo-dup:

No

Feature

We may hide the injections displayed with CSS, but then the holes through the rings due to the edges are still there

We may completely hide the injections with CSS + calling redrawVoltageLevelNodes public function

- [ ] Yes
- [x] No


flo-dup:
The demo needs to be enriched with a nad where some injections are slicing some rings to see and test this feature

PR #231, opened 2025-09-24
Handle new converter station based hvdc edge types

ayolab:

- [x] Yes
- [ ] No

- [x] The migration steps are described in the following section

edge type HVDC_LINE removed.
users must handle HVDC_LINE_VSC and HVDC_LINE_LCC instead of the former generic type.


flo-dup (approved):

flo-dup:
The rating is due to lines not covered by current PR

PR #257, opened 2025-11-05
Fix setBranchStates including parallel edges case

CBiasuzzi:

no

Bug fix

setBranchStates: in case of parallel edges, getHalfEdges is not called  with the correct  iEdge and nbGroupedEdges parameters

detects if edge has parallel edges and call this.getHalfEdge, with the correct iEdge and nbGroupedEdges

- [ ] Yes
- [x] No


flo-dup (changes requested):
Could you try fixing the map? But we could live with the concatenating as a first step. Indeed, I think the best way to solve this is to have the index in the metadata produced by powsybl-diagram (and thus in BranchEdge class)

PR #273, opened 2025-11-24
Bump Vite to v7.2.4 to avoid vulnerabilities

flo-dup:

No

Dependencies update

typescript 5.3.3 and 6 vulnerabilities (moderate)

typescript 5.6.3 and 0 known vulnerabitilites

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,7 @@
                 "@types/node": "^22.13.4",
                 "@types/react": "^19.2.2",
                 "@types/react-dom": "^19.2.1",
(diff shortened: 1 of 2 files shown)


rolnico:
Why specifically TypeScript 5.6.3 and not a more recent version? Are there limitations due to some libraries we use?

flo-dup:
> Why specifically TypeScript 5.6.3 and not a more recent version? Are there limitations due to some libraries we use?

I tried to take the minimum version which solved all the vulnerabilities at that time.

PR #278, opened 2025-12-01
Create SVG nodes from diagram metadata in NAD viewer

massimo-ferraro:

no

feature

The NAD viewer shows the SVG provided as input

An SVG writer has been added, that creates the SVG, with nodes, starting from the diagram metadata. The SVG viewer uses the SVG writer to create the SVG to show, if the SVG provided as input is empty

- [ ] Yes
- [X] No


src/components/network-area-diagram-viewer/metadata-utils.ts
```diff
@@ -333,3 +333,8 @@ export function getEdgeType(edge: EdgeMetadata): EdgeType {
 export function getStringEdgeType(edge: EdgeMetadata): string {
     return EdgeType[getEdgeType(edge)];
 }
+
+export function getBusNodesMetadata(nodeId: string, busNodes: BusNodeMetadata[]): BusNodeMetadata[] {
+    const buses = busNodes.filter((bus) => bus.vlNode === nodeId);
```
flo-dup: As we're going through all metadata, it might be worth to avoid filtering for each vlNode to loop over the bus node metadata only once.

PR #312, opened 2026-01-29
Isolate non-react code into a new sub-package

### LINKED ISSUE #311: Isolate strict dependencies to react libraries into a new sub-package

Require React libraries as peer dependencies, even for applications that are not React-based or that do not use the React parts of the library.

Extract and isolate non-React code into a new package within the library. This package could be used by applications without having to include React dependencies. For those who want to use the React code, it would still be possible to rely on the current library dependency as a backward-compatibility feature.

### Describe the motivation

Working on an Angular app, that is forced to import react dependencies, it seems not to be the best way to use the library.

### Extra Information

This means moving NetworkAreaDiagramViewer and SingleLineDiagramViewer into a new package, tentatively called powsybl-network-viewer-core, which would not contain any React dependencies. This new package would then be imported by the powsybl-network-viewer package.

rpa-dacosta:

- [x] Docs have been added / updated (for bug fixes / features)
- [x] A PR or issue has been opened in all impacted repositories (if any)

Fixes #311 

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
         "check:pack": "attw --pack",
         "build": "vite build",
         "build:watch": "vite build --watch --mode development",
-        "build:all": "npm run build -w @powsybl/network-map-layers && npm run build -w @powsybl/network-viewer-core && npm run build",
+        "build:all": "npm run build --workspaces && npm run build",
         "prepublishOnly": "$npm_execpath run build",
         "test": "jest",
         "test:coverage": "jest --coverage",
--- a/packages/network-viewer-core/package.json
+++ b/packages/network-viewer-core/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@powsybl/network-viewer-core",
-    "version": "3.1.0",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
         "check:pack": "attw --pack",
         "build": "vite build",
         "build:watch": "vite build --watch --mode development",
-        "build:all": "npm run build -w @powsybl/network-map-layers && npm run build",
+        "build:all": "npm run build -w @powsybl/network-map-layers && npm run build -w @powsybl/network-viewer-core && npm run build",
```
amarant: It could be simplified to workspaces build because the order doesn't matter : 

```suggestion
        "build:all": "npm run build --workspaces && npm run build",
```

amarant (dismissed):

package.json
rpa-dacosta: Thanks for the review ! Resolved ✅

PR #313, opened 2026-01-29
[nad-viewer] create middle-edge labels (adaptiveTextZoom)

CBiasuzzi:

no

fix

When the adaptiveTextZoom feature is enabled (enableAdaptiveTextZoom set to true), the edges' infos in the diagram are displayed only when the zoom level is below a threshold;
when above the threshold, edges' infos should be able to be re-drawn from the metadata (and/or from the other existing svg elements).
Currently, the re-drawing of middle-edges labels is not implemented.

This update adds the drawing of middle-edges labels. 

Note: at this stage it is covered only the drawing of a single text middle-edge label; a potential presence of an arrow is not considered.

- [ ] Yes
- [ ] No

**Other information**:


sBouzols:
Tests done in GridSuite project and seems to work great ! 👍
By the way we encounter a limit case where we add in the SVG both extremities edge labels even if only one extremity is visible.
Could we check if added Edge labels are inside the visible area before adding it to the SVG ?

PR #369, opened 2026-04-02
[nad-viewer] Fix Edge Infos mis positioned when using `adaptiveTextZoom` and `moveTextNodeToCoordinates` or `moveNodeToCoordinates`

sBouzols:

No

Bug fix

When using nad-viewer and public functions `moveTextNodeToCoordinates` or `moveNodeToCoordinates` in the client side, with de `adaptiveTextZoom` mode, and a NAD containing half edges, sometimes the half edge Infos are completely mis positioned.

fix(NetworkAreaDiagramViewer): reset `initialPosition` at the end of public functions `moveTextNodeToCoordinates` and `moveNodeToCoordinates` to avoid future mis computation with `getTranslatedPolyline` in the `enableAdaptiveTextZoom` mode and halfEdges.
Suggestion : should we reset `initialPosition` at the beginning of `adaptiveZoomViewboxUpdate` to avoid other scenario of this kind ? @CBiasuzzi ?

- [ ] Yes
- [x] No

### DIFF
--- a/demo/src/diagram-viewers/add-diagrams.ts
+++ b/demo/src/diagram-viewers/add-diagrams.ts
@@ -451,14 +451,16 @@ export const addNadToDemo = () => {
                 onBendLineCallback: handleLineBending,
 
                 enableAdaptiveTextZoom: true,
-                adaptiveTextZoomThreshold: 850,
+                adaptiveTextZoomThreshold: 3000,
             };
-            new NetworkAreaDiagramViewer(
+            const nadViewer = new NetworkAreaDiagramViewer(
                 document.getElementById('svg-container-nad-multibus-vlnodes-limit-percentage')!,
                 svgContent,
                 NadSvgMultibusVLNodesLimitPercentageExampleMeta,
                 nadViewerParametersOptions
             );
+
+            nadViewer.moveNodeToCoordinates('VL5', -100, -50);
         });
 
     fetch(NadSvgMultibusVLNodes14Example)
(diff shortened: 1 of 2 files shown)


CBiasuzzi (commented):
this.InitialPosition is set during a drag & drop action when a node is moved. It stores the node’s previous position while it is being moved and is used to compute the correct positions of the elements that are affected
(Its semantic is  closer to a 'previous position point' than to a real 'initial point',  since it is used to compute a delta).

When the dragging of a node ends (i.e., when the user releases the mouse button), the code in onMouseLeftUpOrLeave resets the InitialPosition value. 
Therefore, this variable is normally null.

When adaptiveZoomViewboxUpdate is enabled and triggered through a zoom or pan, the viewer must rebuild the edge info elements.
To rebuild the edge infos, some of the methods used for drag & drop are involved — e.g., getHalfVisibleHalfEdges (which uses the current InitialPosition to compute positions).

InitialPosition is also set during the API calls moveTextNodeToCoordinates and moveNodeToCoordinates (through the updateElement method).
However, until the fix introduced in this branch, initialPosition had never been reset to null at the end of these API methods. 
As a result,  InitialPosition could contain stale data and cause the problem discussed in this PR (the misplaced labels in the first clip).

During a zoom or pan  interaction, nodes are not normally moved at the same time, so a non-null InitialPosition should not really exist, during the adaptiveZoomViewboxUpdate.
However, as illustrated in the second clip in this PR, while a node is being dragged (without releasing the mouse button), a zoom action can actually  be performed (e.g., using the mouse wheel).
The zoom action activates the adaptiveZoomViewboxUpdate, and in this case, the InitialPosition would not be null.

Overall, resetting the InitialPosition at the start of adaptiveZoomViewboxUpdate seems to be necessary to prevent the issue.
(I wonder if the current InitialPosition could be saved before the reset and restored at the end of adaptiveZoomViewboxUpdate ? just to keep it as non invasive as possible.
Note that during the adaptiveZoomViewboxUpdate, the initialPosition is never set)

What do you think?

CBiasuzzi (approved):
It looks very good to me. I made some tests, too: It worked as expected, also when I was trying to drag a node and zoom the diagram in/out, at the same time.

PR #384, opened 2026-05-04
[nad-viewer] Handle edge info components in node moving and adaptive zoom

massimo-ferraro:

no

feature

In the NAD viewer, the svg components used as edge info are not handled, when moving nodes or when using the adaptive zoom functionality

In the NAD viewer, the svg components used as edge info are handled, when moving nodes or when using the adaptive zoom functionality

- [ ] Yes
- [X] No


packages/network-viewer-core/src/network-area-diagram-viewer/component-utils.ts
```diff
@@ -0,0 +1,31 @@
...
+import LockSvg from '../resources/default-library/lock.svg';
+import FlashSvg from '../resources/default-library/flash.svg';
+import UnknownComponentSvg from '../resources/default-library/unknown-component.svg';
+import { SVG } from '@svgdotjs/svg.js';
+
+const ComponentSvgMapping: { [key: string]: string } = {
+    'lock.svg': LockSvg,
+    'flash.svg': FlashSvg,
+    'unknown-component.svg': UnknownComponentSvg,
+};
```
rolnico:
```suggestion
import { LibraryComponent } from './library-component';
import { SVG } from '@svgdotjs/svg.js';

const svgModules = import.meta.glob('../resources/default-library/*.svg', {
    query: '?url',
    import: 'default',
    eager: true,
});

const ComponentSvgMapping: Record<string, string> = Object.fromEntries(
    Object.entries(svgModules).map(([modulePath, url]) => {
        const fileName = modulePath.split('/').at(-1)!;
        return [fileName, url as string];
    })
);
```

Something like this could allow us to avoid having to declare manually all the components

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -142,6 +145,8 @@ export class NetworkAreaDiagramViewer {
 
     nodeMap: Map<string, NodeMetadata> | null = null;
 
+    componentLibrary: LibraryComponent[] = LibraryComponents;
```
rolnico: It would be nice to have the possibility to use another components library.
Here is a patch that should provide such possibility: 

[Allow_components_library_other_than_the_default_one.patch](https://github.com/user-attachments/files/27355692/Allow_components_library_other_than_the_default_one.patch)

(code might have to be perfected, of course!)

rolnico (changes requested):

packages/network-viewer-core/src/network-area-diagram-viewer/component-utils.ts
massimo-ferraro: Used suggested approach for loading components

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: The patch seems fine, I applied it

rolnico (approved):

antoinebhs (approved):

PR #397, opened 2026-05-29
[v3.4.0] fix middle arrow direction

souissimai:

Bug fix

The middle arrow of an edge is displayed in the wrong direction. The front-end recalculates the arrow rotation every time the diagram is updated (on init, on node move, on branch state update). The rotation was incorrectly computed using the IN/OUT direction from the metadata, but this direction is already encoded in the SVG arrow path shape itself — applying it again in the rotation caused the arrow to flip in certain cases.

The middle arrow rotation is now computed purely from the geometry of the half-edges. A 180° flip is applied only when halfEdge2 is used alone, to compensate for its reversed geometric convention relative to the canonical node1→node2 direction. The IN/OUT direction no longer influences the rotation calculation.

- [ ] Yes
- [ ] No

**Other information**:


rolnico (changes requested):
Same as the other PR, the base branch must be `main`

rolnico (approved):

PR #399, opened 2026-06-01
Fix middle arrow direction

ayolab:
*Please check if the PR fulfills these requirements**

No

Bug fix

The middle arrow of an edge is displayed in the wrong direction. The front-end recalculates the arrow rotation every time the diagram is updated (on init, on node move, on branch state update). The rotation was incorrectly computed using the IN/OUT direction from the metadata, but this direction is already encoded in the SVG arrow path shape itself — applying it again in the rotation caused the arrow to flip in certain cases.

The middle arrow rotation is now computed purely from the geometry of the half-edges. A 180° flip is applied only when halfEdge2 is used alone, to compensate for its reversed geometric convention relative to the canonical node1→node2 direction. The IN/OUT direction no longer influences the rotation calculation.

- [ ] Yes
- [x] No


packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -1392,9 +1392,9 @@ export class NetworkAreaDiagramViewer {
 
         let factor: number = 1;
         if (direction) {
-            const arrowAngle = HalfEdgeUtils.getMiddleArrowRotation(halfEdge1, halfEdge2, direction);
+            const arrowAngle = HalfEdgeUtils.getMiddleArrowRotation(halfEdge1, halfEdge2);
             const secondArrowAngle: number | undefined = directionA
-                ? HalfEdgeUtils.getMiddleArrowRotation(halfEdge1, halfEdge2, directionA)
+                ? HalfEdgeUtils.getMiddleArrowRotation(halfEdge1, halfEdge2)
```
massimo-ferraro: You could simplify the code, avoiding to have and compute the secondArrowAngle variable, as either it is undefined, or it has the same value of the arrowAngle variable. You could also simplify the redrawArrows method, the secondArrowAngle optional parameter is useless, for the same reason.

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -1363,9 +1363,7 @@ export class NetworkAreaDiagramViewer {
                     'transform',
                     arrowRotateString + ' translate(' + DiagramUtils.getFormattedPoint(new Point(0, -shift)) + ')'
                 );
-            if (secondArrowAngle) {
-                arrowRotateString = 'rotate(' + DiagramUtils.getFormattedValue(secondArrowAngle) + ')';
-            }
+            arrowRotateString = 'rotate(' + DiagramUtils.getFormattedValue(arrowAngle) + ')';
```
massimo-ferraro: I think you could directly remove this line, the same value is assigend to this variable some lines above

ayolab: Done

massimo-ferraro (approved):

PR #400, opened 2026-06-01
Fix NAD side branch arrows direction

ayolab:

No

Bug fix

If `labelB` in edge infos metadata value is positive but `direction="IN"` then the previous behaviour was to override the arrow `direction`

We set the `direction` based on the value sign only if it's absent from the edge infos metadata.

- [ ] Yes
- [x] No


packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -2264,11 +2263,11 @@ export class NetworkAreaDiagramViewer {
...
                 ? value.toFixed(DiagramUtils.getEdgeInfoValuePrecision(edgeInfoMetadata.infoTypeB, this.svgParameters))
                 : value;
-        const direction = typeof value === 'number' ? DiagramUtils.getArrowDirection(value) : undefined;
-        if (edgeInfoMetadata.directionB) {
-            edgeInfoMetadata.directionB = direction;
-        } else {
-            edgeInfoMetadata.direction = direction;
+        // set direction based on the value sign if missing
+        // Do not override direction if present based on value sign
+        // Because it's ok to have a positive value and an IN direction
```
rolnico: Remove comment

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -2264,11 +2263,8 @@ export class NetworkAreaDiagramViewer {
             typeof value === 'number'
                 ? value.toFixed(DiagramUtils.getEdgeInfoValuePrecision(edgeInfoMetadata.infoTypeB, this.svgParameters))
                 : value;
-        const direction = typeof value === 'number' ? DiagramUtils.getArrowDirection(value) : undefined;
-        if (edgeInfoMetadata.directionB) {
-            edgeInfoMetadata.directionB = direction;
-        } else {
-            edgeInfoMetadata.direction = direction;
+        if (!edgeInfoMetadata.directionB && !edgeInfoMetadata.direction) {
+            edgeInfoMetadata.direction = typeof value === 'number' ? DiagramUtils.getArrowDirection(value) : undefined;
```
massimo-ferraro, line 2305: the updateEdgeInfoMetadata function is used by the adaptive zoom and by the setBranchStates functionality.
the setBranchStates allows to set the label value, not the direction (old implementation, never updated with the new edge parameters).
With this change, the setBranchStates functionality would compute the direction the first time you ser the values for an edge, but it would not change the direction in the following changes (see the slider in the Eurostag demo example). Just check this is the expected behaviour, for this functionality (it should be used in some jupyter widgets)

ayolab: good catch, we don't have this behavior in gridsuite
Fixed

massimo-ferraro (approved):

PR #401, opened 2026-06-01
Perf : avoid CSS recalculation on cursor change

ayolab:

No

Create transparent overlay rect for cursor changes to fix the perf issues on cursor changes

Setting cursor on svgDraw.node or drawnSvg.parentElement.style (ancestor of all NAD nodes) triggers a full CSS style recalculation across thousands of descendants (inherited property cascade) causing a multi-second freeze on large diagrams

A transparent overlay rect for cursor changes : a leaf rect with no children to avoid the style recalculation.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
+++ b/packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
@@ -708,6 +708,9 @@ export class NetworkAreaDiagramViewer {
         if (!this.svgDraw) {
             return;
         }
+        // Detach any existing overlay before creating a new one
+        this.detachCursorOverlay();
+
         // Size of the transparent cursor overlay rect. Oversized on purpose so it always covers the viewport at any pan/zoom
         const size = 2000000;
         const overlay = this.svgDraw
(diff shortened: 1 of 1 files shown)


packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -704,12 +699,35 @@ export class NetworkAreaDiagramViewer {
...
+    // Appends a transparent overlay rect on top of svgDraw with the given cursor.
+    // Setting cursor on svgDraw.node or any ancestor of NAD nodes triggers a full CSS style
+    // recalculation across thousands of descendants (inherited property cascade) causing
+    // a multi-second freeze on large diagrams. Using a leaf rect with no children avoids this.
+    // The rect is created on demand so it is absent from the DOM (and from SVG exports) at rest.
+    private attachCursorOverlay(cursor: string) {
+        if (!this.svgDraw) {
+            return;
+        }
+        // Size of the transparent cursor overlay rect. Oversized on purpose so it always covers the viewport at any pan/zoom
```
rolnico, line 714:
```suggestion
        // Detach any existing overlay before creating a new one
        this.detachCursorOverlay();

        // Size of the transparent cursor overlay rect. Oversized on purpose so it always covers the viewport at any pan/zoom
```
If `attachCursorOverlay` is called twice before `detachCursorOverlay` (e.g., a theoretically possible edge case where events fire out of order), `this.cursorOverlayRect` would be overwritten and the first rect would leak in the DOM. A defensive `detachCursorOverlay()` at the top of `attachCursorOverlay` would prevent this.

rolnico (changes requested):

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
ayolab: Done

rolnico (approved):

PR #417, opened 2026-06-18
fixes redrawTextEdge when enableLevelOfDetail is enabled

### LINKED ISSUE #162: Moving of text nodes in NAD viewer with enableLevelOfDetail = true

In the NAD viewer, with enableLevelOfDetail = true and the text nodes are not visualized, when moving a voltage level node, the text edges are not correctly moved

In the NAD viewer, with enableLevelOfDetail = true and the text nodes are not visualized, when moving a voltage level node, the text edges are correctly moved

### Describe the steps

In the NAD viewer, with enableLevelOfDetail = true

Image

Zooming out so that the text nodes are not visualized

Image

Then moving a voltage level node

Image

And zooming in, so that the text nodes are again visualized

Image

the text node is correctly moved, along the voltage level node, but the text edge is not, connecting to the wrong point.
Moving again the voltage level node, or the text node, when the text node is visualized, fixes the connection.

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

The viewer, when moving text edges, gets the size of the text nodes using scrollHeight and scrollWidth properties of the div inside the foreignObject of the text node. The size is used for correctly move the text edges, attaching it to the proper points.
When enableLevelOfDetail = true and text nodes are not visualized, both scrollHeight and scrollWidth preoperties are 0.
A recent hotfix, put for correctly moving text nodes programmatically (see PR #126), does not compute the end of text edges if scrollHeight and scrollWidth preoperties are 0 (they are 0 when moving text nodes programmatically, in these cases the end of text edges is computed before, and the text edge is connected correctly). 
So, when the size is 0 because the thext node are not visualized, the  the text edge is connected to the wrong point: 0,0, as in the example above, or to the previous computed text edge end, if a movement was done with the text nodes visualized.
Even changing the code, for distinguishing if the size is 0,0 because the text node is moved programmatically or not, and so computing the text edge when needed, the viewer would possibly connect the edge to the wrong point, when the connection is not the default one, see below

Image

CBiasuzzi:

fixes #162

- [ ] Yes
- [ ] No

**Other information**:

### DIFF
--- a/packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
+++ b/packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
@@ -1114,8 +1114,8 @@ export class NetworkAreaDiagramViewer {
                 // textNode?.firstElementChild?.scrollHeight and textNode?.firstElementChild?.scrollWidth seem not defined,
                 // so textHeight and textWidth are zero; in these cases, compute endTextEdge using metadata
                 this.endTextEdge = new Point(
-                    node.x + textNodeMetadata.connectionShiftX,
-                    node.y + textNodeMetadata.connectionShiftY
+                    node.x + (textNodeMetadata.connectionShiftX ?? 0),
+                    node.y + (textNodeMetadata.connectionShiftY ?? 0)
                 );
             }
             const startTextEdge = DiagramUtils.getPointAtDistance(
--- a/packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
+++ b/packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
@@ -1095,10 +1095,7 @@ export class NetworkAreaDiagramViewer {
             );
             // compute text edge start and end
             const vlNodePosition = new Point(node.x, node.y);
-            // HOTFIX If we call moveElement programmatically (not during a drag and drop event)
-            // then textNode?.firstElementChild?.scrollHeight and textNode?.firstElementChild?.scrollWidth seems not defined
-            // then textHeight and textWidth equal 0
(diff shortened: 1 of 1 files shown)


packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -1107,6 +1104,19 @@ export class NetworkAreaDiagramViewer {
...
+                    textNodeMetadata.connectionShiftX = this.endTextEdge.x - node.x;
+                    textNodeMetadata.connectionShiftY = this.endTextEdge.y - node.y;
+                }
+            } else if (textNodeMetadata) {
+                // when the text node is moved programmatically or hidden by CSS (enableLevelOfDetail feature)
+                // textNode?.firstElementChild?.scrollHeight and textNode?.firstElementChild?.scrollWidth seem not defined,
+                // so textHeight and textWidth are zero; in these cases, compute endTextEdge using metadata
+                this.endTextEdge = new Point(
+                    node.x + textNodeMetadata.connectionShiftX,
+                    node.y + textNodeMetadata.connectionShiftY
```
rolnico: Are `connectionShiftX` and `connectionShiftY` always initiated in the metadata? Because if there are undefined at the beginning, we'll get NaN here

rolnico (changes requested):

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
CBiasuzzi: good point, thanks! I'm adding a fallback ?? 0, in case they are initially undefined. Normally, they should be defined (and the code above keeps them updated)

rolnico (approved):

PR #418, opened 2026-06-19
[nad-viewer] Add edge info components in SVG creation from metadata

massimo-ferraro:

no

feature

In the NAD viewer, the SVG writer that creates the SVG starting from diagram metadata does not handle components in edge infos

In the NAD viewer, the SVG writer that creates the SVG starting from diagram metadata handles components in edge infos

- [X] Yes
- [ ] No

The creation of the SVG from metadata  was previously done in a single step: the SVG was created and provided to the viewer, in the constructor, for the parsing and visualization.
The creation  is now done in 2 steps: 
- the first step creates an empty SVG, with the correct viewbox, to be provided to the viewer for the parsing
- the second step adds all the content to the DOM element handled by the viewer

So the SVG writer methods has been changed for implementing these 2 steps, even if the old single step method is still available.
This changes were done for handling the asynchronous loading of the components.
A new parameter for choosing the creation of the SVG from the diagram metadata has been added to the NAD parameters

### DIFF
--- a/demo/src/nad-created.ts
+++ b/demo/src/nad-created.ts
@@ -26,6 +26,8 @@ import NadSvgPegaseNetworkExample from './diagram-viewers/data/case1354pegase.sv
 import NadSvgPegaseNetworkExampleMeta from './diagram-viewers/data/case1354pegase_metadata.json';
 import NadSvgDoubleArrowsExample from './diagram-viewers/data/nad-double-arrows-with-middle-values.svg';
 import NadSvgDoubleArrowsExampleMeta from './diagram-viewers/data/nad-double-arrows-with-middle-values_metadata.json';
+import NadSvgComponentsExample from './diagram-viewers/data/nad-edge-info-components.svg';
+import NadSvgComponentsExampleMeta from './diagram-viewers/data/nad-edge-info-components_metadata.json';
 
 import { NadViewerParametersOptions, NetworkAreaDiagramViewer } from '../../src';
 import {
@@ -68,6 +70,7 @@ const addCreatedNadToDemo = () => {
         onToggleHoverCallback: handleToggleNadHover,
         onRightClickCallback: handleRightClick,
         onBendLineCallback: handleLineBending,
+        createSvgFromMetadata: true,
     };
     new NetworkAreaDiagramViewer(
         document.getElementById('svg-container-nad-eurostag-c')!,
@@ -304,6 +307,35 @@ const addCreatedNadToDemo = () => {
         nadViewerParametersOptions
     );
 
+    fetch(NadSvgComponentsExample)
+        .then((response) => response.text())
+        .then((svgContent) => {
+            const nadViewerParametersOptions: NadViewerParametersOptions = {
+                enableDragInteraction: true,
+                addButtons: true,
+                onMoveNodeCallback: handleNodeMove,
+                onMoveTextNodeCallback: handleTextNodeMove,
+                onSelectNodeCallback: handleNodeSelect,
+                onToggleHoverCallback: handleToggleNadHover,
+                onRightClickCallback: handleRightClick,
+                onBendLineCallback: handleLineBending,
+                adaptiveTextZoom: { enabled: true, threshold: 1100 },
+            };
+            new NetworkAreaDiagramViewer(
+                document.getElementById('svg-container-nad-components')!,
+                svgContent,
+                NadSvgComponentsExampleMeta,
+                nadViewerParametersOptions
+            );
+        });
+
+    new NetworkAreaDiagramViewer(
+        document.getElementById('svg-container-nad-components-c')!,
+        '',
+        NadSvgComponentsExampleMeta,
+        nadViewerParametersOptions
+    );
+
     fetch(NadSvgPegaseNetworkExample)
         .then((response) => response.text())
         .then((svgContent) => {
@@ -336,6 +368,7 @@ const addCreatedNadToDemo = () => {
         onToggleHoverCallback: handleToggleNadHover,
         onRightClickCallback: handleRightClick,
         onBendLineCallback: handleLineBending,
+        createSvgFromMetadata: true,
     };
     new NetworkAreaDiagramViewer(
         document.getElementById('svg-container-nad-pegase-network-c')!,
--- a/packages/network-viewer-core/src/network-area-diagram-viewer/nad-viewer-parameters.ts
+++ b/packages/network-viewer-core/src/network-area-diagram-viewer/nad-viewer-parameters.ts
@@ -135,6 +135,9 @@ export interface NadViewerParametersOptions {
     // Resolver that maps an SVG filename to its URL, for use with a custom component library.
     // If not provided, the default library SVG files are used.
     svgUrlResolver?: (fileName: string) => string;
+
+    // Whether create the SVG from diagram metadata, instead of using the SVG content provided as input
+    createSvgFromMetadata?: boolean;
 }
 
 export class NadViewerParameters {
@@ -149,6 +152,7 @@ export class NadViewerParameters {
     static readonly HOVER_POSITION_PRECISION_DEFAULT = 10;
     static readonly ENABLE_ADAPTIVE_ZOOM_DEFAULT = false;
     static readonly THRESHOLD_ADAPTIVE_ZOOM_DEFAULT = 3000;
+    static readonly CREATE_SVG_FROM_METADATA_DEFAULT = false;
 
     nadViewerParametersOptions: NadViewerParametersOptions | undefined;
 
@@ -231,5 +235,12 @@ export class NadViewerParameters {
 
     public getSvgUrlResolver(): ((fileName: string) => string) | undefined {
         return this.nadViewerParametersOptions?.svgUrlResolver;
+    }
+
+    public getCreateSvgFromMetadata(): boolean {
+        return (
+            this.nadViewerParametersOptions?.createSvgFromMetadata ??
+            NadViewerParameters.CREATE_SVG_FROM_METADATA_DEFAULT
+        );
     }
 }
--- a/packages/network-viewer-core/src/network-area-diagram-viewer/nad-viewer-parameters.ts
+++ b/packages/network-viewer-core/src/network-area-diagram-viewer/nad-viewer-parameters.ts
@@ -136,7 +136,7 @@ export interface NadViewerParametersOptions {
     // If not provided, the default library SVG files are used.
     svgUrlResolver?: (fileName: string) => string;
 
-    // Whether create the SVG from diagram metadata, instead of using the SVG content provided as input
+    // Whether to create the SVG from diagram metadata, instead of using the SVG content provided as input
     createSvgFromMetadata?: boolean;
 }
 
--- a/packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
+++ b/packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
@@ -150,6 +150,8 @@ export class NetworkAreaDiagramViewer {
 
     componentLibrary: LibraryComponent[] = DefaultLibraryComponents;
 
+    svgWriter: SvgWriter | undefined = undefined;
+
     static readonly ZOOM_CLASS_PREFIX = 'nad-zoom-';
 
     /**
@@ -169,11 +171,11 @@ export class NetworkAreaDiagramViewer {
         this.svgDiv.id = 'svg-container';
         this.svgContent = svgContent;
         this.diagramMetadata = diagramMetadata;
-        if (this.diagramMetadata != null && this.svgContent.length == 0) {
-            const createdSvg = new SvgWriter(this.diagramMetadata).getSvg();
-            this.svgContent = createdSvg;
-        }
         this.nadViewerParameters = new NadViewerParameters(nadViewerParametersOptions ?? undefined);
+        if (this.nadViewerParameters.getCreateSvgFromMetadata() && this.diagramMetadata != null) {
+            this.svgWriter = new SvgWriter(this.diagramMetadata);
+            this.svgContent = this.svgWriter.getEmptySvg();
+        }
         this.width = 0;
         this.height = 0;
         this.originalWidth = 0;
@@ -378,6 +380,9 @@ export class NetworkAreaDiagramViewer {
         };
         this.svgDraw = SVG().addTo(this.svgDiv).size(this.width, this.height).viewbox(viewBox);
         this.innerSvg = <SVGElement>this.svgDraw.svg(this.svgContent).node.firstElementChild;
+        if (this.nadViewerParameters.getCreateSvgFromMetadata()) {
+            this.svgWriter?.addSvgContent(<SVGSVGElement>this.innerSvg);
+        }
         this.innerSvg.style.overflow = 'visible';
 
         this.textNodesSection = this.getOrCreateTextNodesSection();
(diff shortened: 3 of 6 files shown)


packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -378,6 +380,9 @@ export class NetworkAreaDiagramViewer {
         };
         this.svgDraw = SVG().addTo(this.svgDiv).size(this.width, this.height).viewbox(viewBox);
         this.innerSvg = <SVGElement>this.svgDraw.svg(this.svgContent).node.firstElementChild;
+        if (this.nadViewerParameters.getCreateSvgFromMetadata()) {
+            this.svgWriter?.addSvgContent(<SVGSVGElement>this.innerSvg);
```
rolnico:
```suggestion
            if (!this.svgWriter) {
                console.warn(
                    '[NAD] createSvgFromMetadata is true but no SvgWriter was created (diagramMetadata may be null).'
                );
            }
            this.svgWriter?.addSvgContent(<SVGSVGElement>this.innerSvg);
```
Or maybe add the check on the metadata?

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -487,6 +450,53 @@ export class NetworkAreaDiagramViewer {
         }
     }
 
+    private addEvents(svgDraw: Svg, drawnSvg: SVGElement, hasMetadata: boolean) {
```
rolnico: `drawnSvg` is not used in the method?

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
rolnico: Maybe also add a check on the cast?

```suggestion
            const svgEl = this.innerSvg;
            if (!(svgEl instanceof SVGSVGElement)) {
                console.error('[NAD] Expected SVGSVGElement for addSvgContent, got:', svgEl);
                return;
            }
            this.svgWriter?.addSvgContent(svgEl);
```

packages/network-viewer-core/src/network-area-diagram-viewer/nad-viewer-parameters.ts
```diff
@@ -135,6 +135,9 @@ export interface NadViewerParametersOptions {
     // Resolver that maps an SVG filename to its URL, for use with a custom component library.
     // If not provided, the default library SVG files are used.
     svgUrlResolver?: (fileName: string) => string;
+
+    // Whether create the SVG from diagram metadata, instead of using the SVG content provided as input
```
rolnico:
```suggestion
    // Whether to create the SVG from diagram metadata, instead of using the SVG content provided as input
```

demo/src/nad-created.ts
```diff
@@ -304,6 +307,35 @@ const addCreatedNadToDemo = () => {
         nadViewerParametersOptions
     );
 
+    fetch(NadSvgComponentsExample)
+        .then((response) => response.text())
+        .then((svgContent) => {
+            const nadViewerParametersOptions: NadViewerParametersOptions = {
```
rolnico, line 315: Missing `createSvgFromMetadata: true`?

rolnico (changes requested):

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: Check on svgWriter added

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: You are right, it was used before aligning the branch to the main branch. Removed

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: Check added

packages/network-viewer-core/src/network-area-diagram-viewer/nad-viewer-parameters.ts
massimo-ferraro: Fixed

demo/src/nad-created.ts
massimo-ferraro: This viewer is not the one with the SVG content created from metadata. The viewer with SVG from metadata is created some lines of code below, and it has createSvgFromMetadata paramenter true

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -169,11 +171,11 @@ export class NetworkAreaDiagramViewer {
         this.svgDiv.id = 'svg-container';
         this.svgContent = svgContent;
         this.diagramMetadata = diagramMetadata;
-        if (this.diagramMetadata != null && this.svgContent.length == 0) {
-            const createdSvg = new SvgWriter(this.diagramMetadata).getSvg();
-            this.svgContent = createdSvg;
-        }
         this.nadViewerParameters = new NadViewerParameters(nadViewerParametersOptions ?? undefined);
+        if (this.nadViewerParameters.getCreateSvgFromMetadata() && this.diagramMetadata != null) {
+            this.svgWriter = new SvgWriter(this.diagramMetadata);
```
rolnico:
```suggestion
        this.componentLibrary = this.nadViewerParameters.getComponentLibrary() ?? DefaultLibraryComponents;
        if (this.nadViewerParameters.getCreateSvgFromMetadata() && this.diagramMetadata != null) {
            this.svgWriter = new SvgWriter(
                this.diagramMetadata,
                this.componentLibrary,
                this.nadViewerParameters.getSvgUrlResolver()
            );
```

Otherwise the SvgWriter always use the default values (library + svg url resolver)

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
rolnico:
```suggestion
```

Moved up

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -378,6 +380,9 @@ export class NetworkAreaDiagramViewer {
         };
         this.svgDraw = SVG().addTo(this.svgDiv).size(this.width, this.height).viewbox(viewBox);
         this.innerSvg = <SVGElement>this.svgDraw.svg(this.svgContent).node.firstElementChild;
+        if (this.nadViewerParameters.getCreateSvgFromMetadata()) {
+            this.addSvgContent(this.svgWriter, this.innerSvg);
```
rolnico:
```suggestion
        if (this.svgWriter != null) {
            this.svgWriter.addSvgContent(<SVGSVGElement>this.innerSvg);
```

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -487,6 +450,67 @@ export class NetworkAreaDiagramViewer {
...
+                '[NAD] createSvgFromMetadata is true but no SvgWriter was created (diagramMetadata may be null).'
+            );
+            return;
+        }
+        if (!(svgElement instanceof SVGSVGElement)) {
+            console.error('[NAD] Expected SVGSVGElement for addSvgContent, got:', svgElement);
+            return;
+        }
+        svgWriter.addSvgContent(svgElement);
+    }
```
rolnico: This method could be removed

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
```diff
@@ -44,20 +47,47 @@ export class SvgWriter {
...
+        this.addSvgContent(baseSvg.svg);
+        return new XMLSerializer().serializeToString(baseSvg.xmlDoc);
+    }
+
+    public getEmptySvg(): string {
+        const baseSvg = this.createBaseSvg();
+        return new XMLSerializer().serializeToString(baseSvg.xmlDoc);
+    }
+
+    public createBaseSvg(textBoxSize?: { width: number; height: number }): { xmlDoc: XMLDocument; svg: SVGSVGElement } {
```
rolnico:
```suggestion
    private createBaseSvg(textBoxSize?: { width: number; height: number }): {
        xmlDoc: XMLDocument;
        svg: SVGSVGElement;
    } {
```

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
```diff
@@ -44,20 +47,47 @@ export class SvgWriter {
...
-        // get edge router, for computing edges points
-        this.edgeRouter = new EdgeRouter(this.diagramMetadata);
+        const baseSvg = this.createBaseSvg(textBoxSize);
+        this.addSvgContent(baseSvg.svg);
+        return new XMLSerializer().serializeToString(baseSvg.xmlDoc);
+    }
+
+    public getEmptySvg(): string {
+        const baseSvg = this.createBaseSvg();
+        return new XMLSerializer().serializeToString(baseSvg.xmlDoc);
```
rolnico:
```suggestion
        const vb = MetadataUtils.getViewBox(
            this.diagramMetadata.nodes,
            this.diagramMetadata.textNodes,
            this.svgParameters
        );
        return `<svg xmlns="http://www.w3.org/2000/svg" viewBox="${vb.x} ${vb.y} ${vb.width} ${vb.height}"></svg>`;
```

The serialization/deserialisation can probably be avoided

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
```diff
@@ -44,20 +47,47 @@ export class SvgWriter {
         2: 'TWO',
         3: 'THREE',
     };
+    componentLibrary: LibraryComponent[] = DefaultLibraryComponents;
+    urlResolver: ((fileName: string) => string) | undefined = undefined;
```
rolnico, line 51:
```suggestion
    urlResolver: ((fileName: string) => string) | undefined = undefined;
    private abortController: AbortController | undefined = undefined;
```
add `abortController` so that each `addSvgContent` call aborts the previous contrtoller, cancelling in-flight detchs from any prior `init()` invocation

rolnico: The getComponentPath method in component-utils.ts should change to:


```ts
export async function getComponentPath(
    componentFilename: string,
    svgUrlResolver?: (fileName: string) => string,
    signal?: AbortSignal
): Promise<SVGPathElement> {
    const url = svgUrlResolver ? svgUrlResolver(componentFilename) : DefaultComponentSvgMapping[componentFilename];
    if (!url) throw new Error(`Unknown component file: ${componentFilename}`);
    const response = await fetch(url, { signal });
    const svgContent = await response.text();
    const path = SVG().svg(svgContent).node.firstElementChild?.firstElementChild;
    if (!path) throw new Error(`No path found in component SVG: ${componentFilename}`);
    return path as SVGPathElement;
}
```

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
```diff
@@ -44,20 +47,47 @@ export class SvgWriter {
...
+    public createBaseSvg(textBoxSize?: { width: number; height: number }): { xmlDoc: XMLDocument; svg: SVGSVGElement } {
         // create XML doc
         const xmlDoc = this.getXmlDoc();
         // add SVG root element
         const svg = this.getSvgRootElement(textBoxSize);
         xmlDoc.appendChild(svg);
+        return { xmlDoc, svg };
+    }
+
+    public addSvgContent(svg: SVGSVGElement) {
```
rolnico, line 97:
```suggestion
    public addSvgContent(svg: SVGSVGElement) {
        this.abortController?.abort();
        this.abortController = new AbortController();
        this.threeWindingsTransformers = [];
        this.threeWindingsTransformerEdges = [];
```

reset the arrays to not duplicate 3WT nodes

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
```diff
@@ -495,6 +528,22 @@ export class SvgWriter {
...
+        const component = ComponentUtils.getComponent(this.componentLibrary, componentType);
+        const edgeInfoComponent = document.createElementNS('http://www.w3.org/2000/svg', 'g');
+        if (component) {
+            const trans = new Point(-component.size.width / 2, -component.size.height / 2);
+            edgeInfoComponent.setAttribute('transform', 'translate(' + DiagramUtils.getFormattedPoint(trans) + ')');
+            edgeInfoComponent.classList.add(component.styleClass);
+            component.subComponents.forEach((subComponent) => {
+                void ComponentUtils.getComponentPath(subComponent.fileName, this.urlResolver).then((path) => {
+                    edgeInfoComponent.appendChild(path);
+                });
```
rolnico, line 556:
```suggestion
                ComponentUtils.getComponentPath(subComponent.fileName, this.urlResolver, this.abortController?.signal)
                    .then((path) => {
                        edgeInfoComponent.appendChild(path);
                    })
                    .catch((err: unknown) => {
                        if (err instanceof DOMException && err.name === 'AbortError') return;
                        console.error('[NAD] Failed to load component', subComponent.fileName, err);
                    });
```

AbortError is silently ignored while other errors log to console

rolnico (changes requested):

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: fixed

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: done

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: method removed

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
massimo-ferraro: fixed

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
massimo-ferraro: you are right, used suggested code

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
massimo-ferraro: As far as I understood, the `AbortController` is used to correctly handle multiple calls to the `addSvgContent` method. This method is called in the `init` method of the viewer, in turn called by the constructor. So there should not be multiple calls to the `addSvgContent` method, at least not for the same `SvgWriter`. Should be the controller added if, for some reason, the call to the `addSvgContent` method is put also in other point of the viewer code? Take also into account that the `addSvgContent` method, right now, assumes that the input SVG is empty, otherwise the resulting svg would have duplicated elements, if you call it multiple times. So, if you want, I can anyway add the controller (it should not create issues) but I'm not sure it is really necessary and useful here.

PR #420, opened 2026-06-26
Add voltage-level class to edge infos from metadata

flomillot:
## Context

Edge infos (arrows and value labels) live in a flat `nad-edge-infos` group and, unlike voltage level nodes and their labels, do not carry the voltage-level CSS class of the side they belong to. As a result, CSS rules that hide a voltage level (voltage-level filtering, zoom-based hiding) leave these arrows and labels visible.

## Change

The voltage-level class(es) are carried by the edge metadata (computed server-side in powsybl-diagram via `classes1`/`classes2`), so they are applied directly on the edge info SVG element :

- Each side edge info is tagged with only its own side's classes (`classes1` for side 1, `classes2` for side 2); the middle edge info, which belongs to both sides, carries the union of both.
- The class list is built by the caller and passed to `getOrCreateEdgeInfo`, covering both the initial server-rendered infos and those recreated during adaptive zoom.

### DIFF
--- a/packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
+++ b/packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
@@ -379,6 +379,9 @@ export class NetworkAreaDiagramViewer {
         this.textEdgesSection = this.getOrCreateTextEdgesSection();
         this.edgeInfosSection = this.getOrCreateEdgeInfosSection();
 
+        // Tag the server-rendered edge infos with their side's voltage-level class
+        this.addClassesToEdgeInfos();
+
         // add events
         const hasMetadata = this.diagramMetadata !== null;
         if (this.hasNodeInteraction() && hasMetadata) {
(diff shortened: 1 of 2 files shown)


packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -379,6 +379,9 @@ export class NetworkAreaDiagramViewer {
         this.textEdgesSection = this.getOrCreateTextEdgesSection();
         this.edgeInfosSection = this.getOrCreateEdgeInfosSection();
 
+        // Tag the server-rendered edge infos with their side's voltage-level class
+        this.addClassesToEdgeInfos();
+
```
sBouzols: If it's manage through `getOrCreateEdgeInfo` why adding classes here at init ?
It should be clearly conditional to `hasMetadata` I think and there is no other cases where we update the classes of all the SVG from metadata like this, this could be a performance issue I think

flomillot: Fixed as discussed

Meklo (dismissed):

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -2411,7 +2412,9 @@ export class NetworkAreaDiagramViewer {
             edge.edgeInfoMiddle = edgeInfoMetadata;
         }
 
-        const edgeInfo = this.getOrCreateEdgeInfo(edgeInfoMetadata);
+        // the middle edge info belongs to both sides, so it carries the union of their classes
+        const classes = [...new Set([...(edge.classes1 ?? []), ...(edge.classes2 ?? [])])];
```
sBouzols: Actually no need to pass throught a Set to avoid duplicates.
`classList.add` already omit already present classes : 
cf : https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/add 
My bad !

sBouzols (dismissed):
Code review OK (a minor remark)
Test OK
Console warning check OK

sBouzols (approved):

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -2411,7 +2412,9 @@ export class NetworkAreaDiagramViewer {
             edge.edgeInfoMiddle = edgeInfoMetadata;
         }
 
-        const edgeInfo = this.getOrCreateEdgeInfo(edgeInfoMetadata);
+        // the middle edge info belongs to both sides, so it carries the classes of both
```
So-Fras, line 2422: What happens with the middle edge ingo if we have different nominal voltages from one side to another (it can happen unfortunately)?

So-Fras (changes requested):
Thank you for this pull request! I think we still need to clarify a point there (see comment).

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
flomillot: It merges all the classes so it will carry both voltages. Then the consumer decides what to do with it. I dont see any other solution.

So-Fras (approved):

PR #421, opened 2026-06-30
[nad-viewer] Add diagrams to test data for creation of SVG from metadata

massimo-ferraro:

- [X] Tests for the changes have been added (for bug fixes / features)

no

tests update

There are a couple of tests for the creation of NAD SVG from diagram metadata

There are more tests and test data for the creation of NAD SVG from diagram metadata

- [ ] Yes
- [X] No

### DIFF
--- a/packages/network-viewer-core/src/network-area-diagram-viewer/edge-router.ts
+++ b/packages/network-viewer-core/src/network-area-diagram-viewer/edge-router.ts
@@ -176,29 +176,29 @@ export class EdgeRouter {
         const node2Angles: number[] = this.nodeAngles[edge.node2] ?? [];
         node2Angles.push(angle2);
         this.nodeAngles[edge.node2] = node2Angles;
-        this.storeEdgeInfos(edge, halfEdges);
+        this.storeEdgeInfos(edge, halfEdges, false);
     }
 
-    private storeEdgeInfos(edge: EdgeMetadata, halfEdges: HalfEdge[] | null[]) {
+    private storeEdgeInfos(edge: EdgeMetadata, halfEdges: HalfEdge[] | null[], isLoop: boolean) {
         if (edge.edgeInfo1 || edge.edgeInfo2) {
-            this.storeEdgeSideData(edge.svgId, halfEdges);
+            this.storeEdgeSideData(edge.svgId, halfEdges, isLoop);
         }
         if (edge.edgeInfoMiddle) {
             this.storeEdgeMiddleData(edge.svgId, halfEdges);
         }
     }
 
-    private storeEdgeSideData(edgeId: string, halfEdges: HalfEdge[] | null[]) {
+    private storeEdgeSideData(edgeId: string, halfEdges: HalfEdge[] | null[], isLoop: boolean) {
         if (!halfEdges[0] || !halfEdges[1]) {
             return;
         }
         this.edgeSideData[edgeId] = [
             [
-                HalfEdgeUtils.getInfoPoint(halfEdges[0], this.svgParameters),
+                isLoop ? halfEdges[0].edgePoints[1] : HalfEdgeUtils.getInfoPoint(halfEdges[0], this.svgParameters),
                 HalfEdgeUtils.getArrowRotation(halfEdges[0]),
             ],
             [
-                HalfEdgeUtils.getInfoPoint(halfEdges[1], this.svgParameters),
+                isLoop ? halfEdges[1].edgePoints[1] : HalfEdgeUtils.getInfoPoint(halfEdges[1], this.svgParameters),
                 HalfEdgeUtils.getArrowRotation(halfEdges[1]),
             ],
         ];
@@ -366,7 +366,7 @@ export class EdgeRouter {
             return;
         }
         this.edgePoints[edge.svgId] = [halfEdges[0].edgePoints, halfEdges[1].edgePoints];
-        this.storeEdgeInfos(edge, halfEdges);
+        this.storeEdgeInfos(edge, halfEdges, true);
     }
 
     private storeThreeWtEdges(edgesMap: Record<string, EdgeMetadata[]>) {
--- a/packages/network-viewer-core/src/network-area-diagram-viewer/edge-router.ts
+++ b/packages/network-viewer-core/src/network-area-diagram-viewer/edge-router.ts
@@ -17,11 +17,17 @@ import { HalfEdge, LabelData } from './diagram-types';
 export class EdgeRouter {
     diagramMetadata: DiagramMetadata;
     svgParameters: SvgParameters;
+    // points of each half edge of edges
     edgePoints: Record<string, [Point[], Point[]]> = {};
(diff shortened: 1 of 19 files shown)


packages/network-viewer-core/src/network-area-diagram-viewer/edge-router.ts
```diff
@@ -176,29 +176,29 @@ export class EdgeRouter {
...
-            this.storeEdgeSideData(edge.svgId, halfEdges);
+            this.storeEdgeSideData(edge.svgId, halfEdges, isLoop);
         }
         if (edge.edgeInfoMiddle) {
             this.storeEdgeMiddleData(edge.svgId, halfEdges);
         }
     }
 
-    private storeEdgeSideData(edgeId: string, halfEdges: HalfEdge[] | null[]) {
+    private storeEdgeSideData(edgeId: string, halfEdges: HalfEdge[] | null[], isLoop: boolean) {
```
NathanDissoubray, line 200: Could you add a description to the function explaining what `isLoop` entails ?

packages/network-viewer-core/src/network-area-diagram-viewer/half-edge-utils.ts
```diff
@@ -278,22 +278,25 @@ export function getHalfEdges(
...
         busOuterRadius: nodeRadius2.busOuterRadius,
         voltageLevelRadius: nodeRadius2.voltageLevelRadius,
         edgeInfoId: edge.edgeInfo2?.svgId,
         edgePoints: edgePoints[1],
     };
-    return [halfEdge1, halfEdge2];
+    return [
+        point1.x == node1.x && point1.y == node1.y ? halfEdge1 : halfEdge2,
+        point2.x == node2.x && point2.y == node2.y ? halfEdge2 : halfEdge1,
+    ];
```
NathanDissoubray: seems like you often need to compare x, y of a point and a node, maybe you could write a function to do that ?

NathanDissoubray (changes requested):

packages/network-viewer-core/src/network-area-diagram-viewer/edge-router.ts
massimo-ferraro: Added descriptions to the data structures and to the method

packages/network-viewer-core/src/network-area-diagram-viewer/half-edge-utils.ts
massimo-ferraro: You are right, added a function

packages/network-viewer-core/src/network-area-diagram-viewer/edge-router.ts
NathanDissoubray: but what is a loop edge ? when you have 2 edges between the same pair of points ?

massimo-ferraro: In a network area diagram a loop edge is an edge that connects 2 bus nodes of the same voltage level node, e.g. in the following image, L7-8-0 and T7-2-0 are loop edges, while L9-8-0 and L7-5-0 are not
loop_edges
It' a term used in the network area diagram projects

packages/network-viewer-core/src/network-area-diagram-viewer/edge-router.ts
```diff
@@ -176,29 +182,32 @@ export class EdgeRouter {
...
         }
         this.edgeSideData[edgeId] = [
             [
-                HalfEdgeUtils.getInfoPoint(halfEdges[0], this.svgParameters),
+                isLoop ? halfEdges[0].edgePoints[1] : HalfEdgeUtils.getInfoPoint(halfEdges[0], this.svgParameters),
                 HalfEdgeUtils.getArrowRotation(halfEdges[0]),
             ],
             [
-                HalfEdgeUtils.getInfoPoint(halfEdges[1], this.svgParameters),
+                isLoop ? halfEdges[1].edgePoints[1] : HalfEdgeUtils.getInfoPoint(halfEdges[1], this.svgParameters),
```
NathanDissoubray, line 210: so the point 1 always corresponds to the middle of the edge that is the concatenation of the two half edges ?

massimo-ferraro: That's the computation of the position of the edge side info, it's different if you are dealing with a loop edge or a normal edge. For the loop edge the edge side info position corresponds with point 1 of the edge points (point 1 of the half edge points), for the other edges HalfEdgeUtils.getInfoPoint function computes the position.
It's the same logic, anche same computation, done in the diagram project, when computing the position of the edge side infos. The SVG writer and edge router in the NAD viewer project, when creating the SVG, should replicate logic and computations done in network area diagram project

packages/network-viewer-core/src/resources/test-data/nad-double-arrows-with-middle-values.svg
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
NathanDissoubray, line 1: Image

is the position correct with the overlap of 300 and 100 ?

massimo-ferraro: If this image is taken from the packages/network-viewer-core/src/resources/test-data/nad-double-arrows-with-middle-values.svg file, this file was generated by the network area diagram. In this PR the file was copied from the demo files, and put in the test data, to compare it with the file generated by the viewer svg writer (but if I open that file, I do not see any overlapping). If the image is got in another way, could you tell me how, so I can check? Thanks

NathanDissoubray: I took the image from the diff of this PR (same filename)

massimo-ferraro: Not sure how you see the overlapping. Anyway, this is an SVG created by the network area diagram, not by code in this project or this PR

NathanDissoubray: I display the svg and there is an overlap. Even if this is generated by diagram, it would be good to open an issue for this case (because I don't think it's supposed to look like that)

massimo-ferraro: I did a bit more of investigation, I can see the overlap if the CSS is applied to the SVG (likely it changes the font size).
This is already done on the diagram side, when creating the svg (the viewer creates a similar svg), so yes, it could be an issue for the diagram. But I'm not sure if and how this could be solved, when creating the SVG, I'm afraid that adding different info elements (double arrows, internal and external labels, side and middle infos) can have this result

NathanDissoubray (approved):

PR #423, opened 2026-07-08
[nad-viewer] Use SVG writer for creating nodes and edges in adaptive zoom functionality

massimo-ferraro:

no

feature

The adaptive zoom functionality handles text boxes and edge infos, removing and redrawing them according to thresholds. It doesn't handle nodes and edges.

The adaptive zoom functionality handles also nodes and edges, removing and redrawing them according to thresholds. The thresholds could also be associated to voltage levels, for removing and redrawing nodes and edges according to their voltage level

- [X] Yes
- [ ] No

The adaptive zoom functionality has been changed for handling nodes and edges. New sections of the `AdaptiveTextZoomOptions` has been added for configuring the handling of nodes and edges, an example of configuration can be found in realgrid.ts file in the demo.
A new parameter for choosing the creation of the SVG from the diagram metadata has also been added to the NAD parameters


flo-dup:
It looks very promising, great work @massimo-ferraro ! :clap:

Some quick comments on this draft, you are probably aware already but here it comes:
- the svgContent should now be optional in the viewer constructor as it's not needed if metadata is here - I think it should be the default way to use the component. We could keep a viewer with a svg without metadata as input, but I doubt it's of any use
- the nad-edge-infos are below the nad-branch-edges in the svg, leading to arrows being below edges 
image

- the text edges should be optional, we should probably add a parameter in SvgParameters in powsybl-diagram for them
- I had the feeling that sometimes the visible viewbox is wrong, or at least the visible elements: I saw some nodes missing at some zoom levels. I can't reproduce it though, did you see something like that? maybe not due to viewbox?
- Should we add a bit of margin when computing the viewbox? for the dragging part, but in fact it's so quick to generate it when releasing the mouse click that it might not be worth it
- The viewer seems to still be dependent on the viewbox of the original svg
    - when I give an empty svg with no viewbox the wheel-zooming is not really working
    - when I give an empty svg with only the viewbox (and no width/height) the dragging shows that the width taken to write the svg is a bigger than what is visible

- about the demo generating java part, I can't launch the generation of the realGrid data, there seems to be a problem when targeting the zip resource with ResourceDataSource. Besides we should avoid adding a binary file to github

But all that seems to be minor bugs, it's a very promising work! :ok_hand: It was the goal of your long work on integration/nad_create_svg and of @CBiasuzzi work on adaptive zoom text and powsybl-diagram metadata, congratulations already for these first results to both of you - can't wait to see that work merged and integrated.

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -2097,31 +2144,121 @@ export class NetworkAreaDiagramViewer {
...
+        if (maxDisplayedSize > maxEdgesThreshold.threshold && maxEdgesThreshold.voltageLevels == undefined) {
+            this.edgesSection?.replaceChildren();
+            containedElementList.edges = [];
+        }
+        if (containedElementList.nodes.length == 0 && containedElementList.edges.length == 0) return;
+        const vlThreshold = DiagramUtils.getVLThreshold(adaptiveTextZoom.nodeThresholds, maxDisplayedSize);
+        const previousVlThreshold = DiagramUtils.getVLThreshold(
+            adaptiveTextZoom.edgeThresholds,
+            this.getPreviousMaxDisplayedSize()
+        );
```
NathanDissoubray: why is `vlThreshold` calculated with `nodeThresholds` but `previousVlThreshold` is calculated with `edgeThreshold` ?

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
```diff
@@ -404,3 +405,50 @@ export function getLabelShiftAndStyle(
...
+    const internalShiftAndStyle = getLabelShiftAndStyle(angle, false, arrowLabelShift);
+    const externalShiftAndStyle = getLabelShiftAndStyle(angle, true, arrowLabelShift);
+    return {
+        angle: radToDeg(textFlipped ? angle - Math.PI : angle),
+        internal: { shift: internalShiftAndStyle[0], style: internalShiftAndStyle[1] },
+        external: { shift: externalShiftAndStyle[0], style: externalShiftAndStyle[1] },
+    };
+}
+
+export function getMaxThreshold(voltageLevelThreshols: VoltageLevelThreshold[]): VoltageLevelThreshold {
```
NathanDissoubray:
```suggestion
export function getMaxThreshold(voltageLevelThresholds: VoltageLevelThreshold[]): VoltageLevelThreshold {
```

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
```diff
@@ -404,3 +405,50 @@ export function getLabelShiftAndStyle(
...
+    return (voltageLevelThreshols.length == 0 ? [{ threshold: 500000 }] : voltageLevelThreshols).reduce(
+        (a, b) => (a.threshold > b.threshold ? a : b),
+        voltageLevelThreshols[0] ?? [{ threshold: 500000 }]
+    );
+}
+
+export function getMinThreshold(voltageLevelThreshols: VoltageLevelThreshold[]): VoltageLevelThreshold {
```
NathanDissoubray:
```suggestion
export function getMinThreshold(voltageLevelThresholds: VoltageLevelThreshold[]): VoltageLevelThreshold {
```

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
```diff
@@ -404,3 +405,50 @@ export function getLabelShiftAndStyle(
...
+    return (voltageLevelThreshols.length == 0 ? [{ threshold: 500000 }] : voltageLevelThreshols).reduce(
+        (a, b) => (a.threshold > b.threshold ? b : a),
+        voltageLevelThreshols[0] ?? [{ threshold: 500000 }]
+    );
+}
+
+export function getVLThreshold(
+    voltageLevelThreshols: VoltageLevelThreshold[],
```
NathanDissoubray:
```suggestion
    voltageLevelThresholds: VoltageLevelThreshold[],
```

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
```diff
@@ -404,3 +405,50 @@ export function getLabelShiftAndStyle(
...
+        internal: { shift: internalShiftAndStyle[0], style: internalShiftAndStyle[1] },
+        external: { shift: externalShiftAndStyle[0], style: externalShiftAndStyle[1] },
+    };
+}
+
+export function getMaxThreshold(voltageLevelThreshols: VoltageLevelThreshold[]): VoltageLevelThreshold {
+    return (voltageLevelThreshols.length == 0 ? [{ threshold: 500000 }] : voltageLevelThreshols).reduce(
+        (a, b) => (a.threshold > b.threshold ? a : b),
+        voltageLevelThreshols[0] ?? [{ threshold: 500000 }]
+    );
```
NathanDissoubray: how about something like that:
```
if (length === 0) {
   return {threshold: 500000}
} else {
   return voltageLevelThresholds.reduce(...)
}
```

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
```diff
@@ -404,3 +405,50 @@ export function getLabelShiftAndStyle(
...
+}
+
+export function getMinThreshold(voltageLevelThreshols: VoltageLevelThreshold[]): VoltageLevelThreshold {
+    return (voltageLevelThreshols.length == 0 ? [{ threshold: 500000 }] : voltageLevelThreshols).reduce(
+        (a, b) => (a.threshold > b.threshold ? b : a),
+        voltageLevelThreshols[0] ?? [{ threshold: 500000 }]
```
NathanDissoubray: you could use the same if-else here

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -2097,31 +2144,121 @@ export class NetworkAreaDiagramViewer {
...
+            });
+            svgWriter.addNodes(<SVGGElement>this.nodesSection!);
+            svgWriter.addEdgesAndInfos(<SVGGElement>this.edgesSection!);
+        }
+    }
+
+    filterNodes(nodes: NodeMetadata[], vlThreshold: VoltageLevelThreshold, maxDisplayedSize: number): NodeMetadata[] {
+        if (vlThreshold.voltageLevels) {
+            // filter nodes metadata, keep nodes not belonging to vl classes
+            nodes = nodes.filter(
```
NathanDissoubray, line 2211: this modifies the nodes that are passed, which forces the caller to keep a copy if they want to keep the original. Is it fine, or should a new array be created ?

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
```diff
@@ -103,8 +132,15 @@ export class SvgWriter {
...
+        return gNodesElement;
+    }
+
+    public addNodes(gNodesElement: SVGGElement) {
+        (this.nodes ?? this.diagramMetadata.nodes).forEach((node) => {
             if (!node.invisible) {
+                const nodeElement = gNodesElement.querySelector(":scope > [id='" + node.svgId + "']") ?? null;
+                if (nodeElement) return;
                 if (MetadataUtils.isThreeWTNode(node)) {
                     this.threeWindingsTransformers.push(node);
```
NathanDissoubray, line 145: 3WT nodes and edges are pushed into something that is never touched by the adaptative zoom in network-area-diagram-viewer.filterNodes / filterEdges
Won't that mean that 3WT are never filtered ?

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -2097,31 +2144,121 @@ export class NetworkAreaDiagramViewer {
...
+                    maxDisplayedSize < vlThreshold.threshold ||
+                    DiagramUtils.intersectionLength(node.classes, vlThreshold.voltageLevels) == 0
+            );
+        }
+        // filter nodes in SVG, possibly remove also nodes to be redrawn, with multiple buses
+        const validNodeIds = new Set(nodes.map((n) => n.svgId));
+        this.nodesSection?.querySelectorAll('g[id]')?.forEach((node) => {
+            if (
+                !validNodeIds.has(node.id) ||
+                (vlThreshold.voltageLevels == undefined && node.querySelectorAll(':scope > path').length > 0)
```
NathanDissoubray: can you explain this line please ? It's not clear to me what it does

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
```diff
@@ -404,3 +405,50 @@ export function getLabelShiftAndStyle(
...
+    const externalShiftAndStyle = getLabelShiftAndStyle(angle, true, arrowLabelShift);
+    return {
+        angle: radToDeg(textFlipped ? angle - Math.PI : angle),
+        internal: { shift: internalShiftAndStyle[0], style: internalShiftAndStyle[1] },
+        external: { shift: externalShiftAndStyle[0], style: externalShiftAndStyle[1] },
+    };
+}
+
+export function getMaxThreshold(voltageLevelThreshols: VoltageLevelThreshold[]): VoltageLevelThreshold {
+    return (voltageLevelThreshols.length == 0 ? [{ threshold: 500000 }] : voltageLevelThreshols).reduce(
```
NathanDissoubray: the 500000 is already defined in nad-viewer-parameters:
```ts
static readonly THRESHOLD_NODES_ADAPTIVE_ZOOM_DEFAULT = 500000;
static readonly THRESHOLD_EDGES_ADAPTIVE_ZOOM_DEFAULT = 500000;
```
Why not use this static value ?

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -169,9 +174,10 @@ export class NetworkAreaDiagramViewer {
...
         this.diagramMetadata = diagramMetadata;
-        if (this.diagramMetadata != null && this.svgContent.length == 0) {
-            const createdSvg = new SvgWriter(this.diagramMetadata).getSvg();
-            this.svgContent = createdSvg;
+        this.nadViewerParameters = new NadViewerParameters(nadViewerParametersOptions ?? undefined);
+        if (this.nadViewerParameters.getCreateSvgFromMetadata() && this.diagramMetadata != null) {
+            this.svgWriter = new SvgWriter({ diagramMetadata: this.diagramMetadata });
+            this.svgContent = this.svgWriter.getEmptySvg();
         }
         this.nadViewerParameters = new NadViewerParameters(nadViewerParametersOptions ?? undefined);
```
NathanDissoubray:
```suggestion
        this.nadViewerParameters = new NadViewerParameters(nadViewerParametersOptions ?? undefined);
        if (this.nadViewerParameters.getCreateSvgFromMetadata() && this.diagramMetadata != null) {
            this.svgWriter = new SvgWriter({ diagramMetadata: this.diagramMetadata });
            this.svgContent = this.svgWriter.getEmptySvg();
        }
```

NathanDissoubray: the second `new NadViewerParameters` overwrites the first one, I don't think this is intended, correct me if I'm wrong

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -2097,31 +2144,121 @@ export class NetworkAreaDiagramViewer {
...
+                diagramMetadata: this.diagramMetadata,
+                elementList: { nodes: nodes, edges: edges },
+                voltageLevels: maxDisplayedSize > vlThreshold.threshold ? vlThreshold.voltageLevels : undefined,
+            });
+            svgWriter.addNodes(<SVGGElement>this.nodesSection!);
+            svgWriter.addEdgesAndInfos(<SVGGElement>this.edgesSection!);
+        }
+    }
+
+    filterNodes(nodes: NodeMetadata[], vlThreshold: VoltageLevelThreshold, maxDisplayedSize: number): NodeMetadata[] {
```
NathanDissoubray: why is this not private ?

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -2097,31 +2144,121 @@ export class NetworkAreaDiagramViewer {
...
+                !validNodeIds.has(node.id) ||
+                (vlThreshold.voltageLevels == undefined && node.querySelectorAll(':scope > path').length > 0)
+            ) {
+                node.remove();
+            }
+        });
+        return nodes;
+    }
+
+    filterEdges(
```
NathanDissoubray: why is this not private ?

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
```diff
@@ -404,3 +405,50 @@ export function getLabelShiftAndStyle(
...
+        if (threshold.threshold < maxDisplayedSize) {
+            return threshold;
+        }
+    }
+    return sortedThresholds.at(-1) ?? getMinThreshold(voltageLevelThreshols);
+}
+
+export function getVLThresholdClasses(voltageLevels: string[]): string {
+    return '.' + voltageLevels.join(',.');
+}
```
NathanDissoubray: I don't see this used anywhere, do you plan to use it for something later ?

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: You are right, I should have used the `edgeThresholds` for the edges filtering, not the `nodeThresholds` (the `previousVlThreshold` is needed only for the edge filtering). Fixed,

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
massimo-ferraro: fixed the typo

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
massimo-ferraro: fixed the typo

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
massimo-ferraro: fixed the typo

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
massimo-ferraro: Changed for using the if else, even if in the short-hand version, but anyway it avoids to reduce an array with only one element

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
massimo-ferraro: done as above

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: Actyually it does not change the original array: the filter function returns the filtered array, but it does not modify it, so the assignement, and the assignment to the input array is an assigment, not a mutation, so it works only on the local variable.

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
massimo-ferraro: The three windings transfomers are not hanled in the adaptive zoom funcionality, so far. They are not removed from the svg (they are in a different svg section), and they would not be redrawn by the svg writer, as you noted. This part still need to be added.

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: This code was added to handle possible nodes redrawing, when a node edges (the edges connected to the node) are removed from the svg, and the node is not. In these cases, if the node has multiple buses, you could need to redraw it. I'm not sure if it would really happen, likely you don't want to remove the edges and keep the node. If this is the case this could be removed. Anyway, thanks for the comment, I found a small fix to do for this line.

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
massimo-ferraro: Now I use the static value

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: You are right, removed the duplicated line

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: Likely created by the IDE, I now made it private

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: Likely created by the IDE, I now made it private

packages/network-viewer-core/src/network-area-diagram-viewer/diagram-utils.ts
massimo-ferraro: You are right, I created and used this method during my developments. In the end I removed the usage, but forgot to remove the function. Removed now

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
NathanDissoubray: Can you add a TODO at the relevant places so others can understand that it's intended that way for now ?

massimo-ferraro:
> It looks very promising, great work @massimo-ferraro ! 👏

Thanks

> 
> Some quick comments on this draft, you are probably aware already but here it comes:
> 
>     * the svgContent should now be optional in the viewer constructor as it's not needed if metadata is here - I think it should be the default way to use the component. We could keep a viewer with a svg without metadata as input, but I doubt it's of any use

Now, in this branch, we have:

- the svg writer, that can create the whole svg, starting from the metadata, and ignoring the svg content provided as input. This operation is done according to nad parameter `createSvgFromMetadata`. This, creating everything in the svg, takes some time to create the whole real grid
- the adaptive zoom functionality, that uses the svg writer to create part of the svg (nodes and edges). This functionality, at least in my tests, starts from the svg content provided as input, removes what needs to be removed and redraws what needs to be redrawn

If you would like to start without the content, and create what is needed in the init, adapting the first situation (the svg writer that creates the whole svg) to work with filtered elements and some voltage levels would require some time and some work.
Another option, quicker to implement, would be to use the adaptive zoom funtionality for this goal: at the init, if the adaptive zoom is enebaled (or should we add a new nad parameter?) you  do not use the input content, create an empty svg (but with the correct viewbox and size, as you noted they are needed), and let the adaptive zoom functionality to fill the svg with the necessary elements (likely as you already did in your tests).
Should we proceed this way?

> 
>     * the nad-edge-infos are below the nad-branch-edges in the svg, leading to arrows being below edges
> 
> 
> image
>

Fixed 

>     * the text edges should be optional, we should probably add a parameter in SvgParameters in powsybl-diagram for them
> 

If this parameter is added, it could be used by the adaptive zoom, and also for the whole SVG creation from metadata

>     * I had the feeling that sometimes the visible viewbox is wrong, or at least the visible elements: I saw some nodes missing at some zoom levels. I can't reproduce it though, did you see something like that? maybe not due to viewbox?
> 

I saw something like that with some combination of `maxWidth` input parameter and browser zoom. I tried to change how the `maxWidth` input parameter is computed (see realgrid.ts), but it could still have happened if you changed the browser zoom after loading the page. I tried changing the css class of the window in the real grid html page, it could have solved the problem (to be tested). btw. this computation of the width could also be put inside the viewer, instead of in the code that calls the creation of the viewer
 
>     * Should we add a bit of margin when computing the viewbox? for the dragging part, but in fact it's so quick to generate it when releasing the mouse click that it might not be worth it
> 
>     * The viewer seems to still be dependent on the viewbox of the original svg
>       
>       * when I give an empty svg with no viewbox the wheel-zooming is not really working
>       * when I give an empty svg with only the viewbox (and no width/height) the dragging shows that the width taken to write the svg is a bigger than what is visible
> 
>     * about the demo generating java part, I can't launch the generation of the realGrid data, there seems to be a problem when targeting the zip resource with ResourceDataSource. Besides we should avoid adding a binary file to github
> 
> 
> But all that seems to be minor bugs, it's a very promising work! 👌 It was the goal of your long work on integration/nad_create_svg and of @CBiasuzzi work on adaptive zoom text and powsybl-diagram metadata, congratulations already for these first results to both of you - can't wait to see that work merged and integrated.

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
massimo-ferraro: Added a TODO comment

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -2097,31 +2143,133 @@ export class NetworkAreaDiagramViewer {
...
+            containedElementList.nodes = [];
+        }
+        const maxEdgesThreshold = DiagramUtils.getMaxThreshold(adaptiveTextZoom.edgeThresholds);
+        if (maxDisplayedSize > maxEdgesThreshold.threshold && maxEdgesThreshold.voltageLevels == undefined) {
+            this.edgesSection?.replaceChildren();
+            containedElementList.edges = [];
+        }
+        if (containedElementList.nodes.length == 0 && containedElementList.edges.length == 0) return;
+        const nodeVlThreshold = DiagramUtils.getVLThreshold(adaptiveTextZoom.nodeThresholds, maxDisplayedSize);
+        const edgeVlThreshold = DiagramUtils.getVLThreshold(adaptiveTextZoom.nodeThresholds, maxDisplayedSize);
```
NathanDissoubray: why use `adaptativeTextZoom.nodeThresholds` to calcule the edge threshold instead of `adaptativeTextZoom.edgeThresholds` ?

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
NathanDissoubray: I looked at the code, doesn't seem fixed to me, I made another comment on it in a new review

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
NathanDissoubray: I don't see the TODO, which line is it on ?

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
NathanDissoubray: what does the querySelectorAll('g[id]') does ? do you think there would be a simpler way to write it, or by naming things in variables before ?

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: You are right, it should be fixed now

packages/network-viewer-core/src/network-area-diagram-viewer/svg-writer.ts
massimo-ferraro: The SVG writer is complete, for what concerns the creation of three windings transformers. It's the NAD viewer that, in the adaptive zoom functionality, does not call the code for creating them (and actually, it does not also contain the code for removing them, if needed). The TODO comment is in the `NetworkAreaDiagramViewer` class file, line 2156

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
massimo-ferraro: fixed now

NathanDissoubray:
I used Arnaud's PR for performance benchmark (#455) to check the performance of this PR compared to the integration branch:
integration on latest commit
```
Interaction  Duration   Frames  Avg FPS  p95 frame  Worst frame  Stutters      Long frames
-----------  ---------  ------  -------  ---------  -----------  ------------  -----------
zoom         8915.1 ms  126     14.1     232.6 ms   449.5 ms     93 (73.8%)    54
pan          6432.6 ms  103     16.0     116.8 ms   150.1 ms     103 (100.0%)  37
moveNode     7149.1 ms  105     14.7     216.7 ms   266.8 ms     103 (98.1%)   28
```

And on this PR:
```
Interaction  Duration    Frames  Avg FPS  p95 frame  Worst frame  Stutters    Long frames
-----------  ----------  ------  -------  ---------  -----------  ----------  -----------
zoom         91567.2 ms  179     2.0      2732.8 ms  15863.9 ms   53 (29.6%)  40
pan          1233.1 ms   69      56.0     16.8 ms    83.4 ms      2 (2.9%)    1
moveNode     2666.8 ms   99      37.1     50.1 ms    66.7 ms      30 (30.3%)  13
```
As we can see, it improves panning and node move (since there are less things to take into account when doing those operations), but it deteriorates zooming.

I think this might be because we need to consider what to draw each time we zoom, but I'm not sure this is the entire reason. A way to solve this might be to not consider edges / nodes that were already not considered before: because we are zooming, if they were not in view, zooming will not bring them in view either.
What do you think @massimo-ferraro ?

PR #428, opened 2026-07-14
[nad-viewer] Add SVG width and height in NAD creation from metadata

massimo-ferraro:

- [X] Tests for the changes have been added (for bug fixes / features)

no

feature

In the NAD viewer, the SVG writer that creates the SVG starting from diagram metadata does not add width and height to the SVG, even if the svgWidthAndHeightAdded metadata parameter is set to true

In the NAD viewer, the SVG writer that creates the SVG starting from diagram metadata, if the svgWidthAndHeightAdded metadata parameter is set to true, adds width and height to the SVG

- [X] Yes
- [ ] No

The PR renamed the metadata utils `getViewBox` function to `getViewBoxAndDimensions`, changing the result type from `ViewBox` to `Dimensions`, i.e. including in the results also the size of the SVG, not only the viewbox. You can see the changes in the function test


demo/java/network-viewer-demo-data/src/main/java/com/powsybl/viewer/demo/NadDemoFiles.java
```diff
@@ -245,6 +245,17 @@ private static NadParameters getNadParametersWithDefaultLabelProvider() {
...
     }
+    
+    private static NadParameters getNadParametersWithSizeAdded() {
+        SvgParameters svgParameters = new SvgParameters()
+            .setCssLocation(SvgParameters.CssLocation.EXTERNAL_NO_IMPORT)
+            .setSvgWidthAndHeightAdded(true)
+            .setFixedHeight(600);
+        return new NadParameters()
+            .setSvgParameters(svgParameters)
+            .setLayoutFactory(new BasicForceLayoutFactory())
```
NathanDissoubray, line 256: any reason to use basic over Atlas2 ?

packages/network-viewer-core/src/network-area-diagram-viewer/metadata-utils.ts
```diff
@@ -221,12 +221,12 @@ export function getRightClickableElementData(
...
 // get view box computed starting from node and text positions
 // defined in diagram metadata
-export function getViewBox(
+export function getViewBoxAndDimensions(
     nodes: NodeMetadata[] | undefined,
     textNodes: TextNodeMetadata[] | undefined,
     svgParameters: SvgParameters,
     textBoxSize?: { width: number; height: number }
-): ViewBox {
+): Dimensions {
```
NathanDissoubray, line 229: this function is exported so this is a breaking change (both the name and the return type change)

packages/network-viewer-core/src/network-area-diagram-viewer/metadata-utils.test.ts
```diff
@@ -321,71 +321,73 @@ test('getViewBox', () => {
...
-    expect(viewBox.y).toBe(-740);
-    expect(viewBox.width).toBe(1700);
-    expect(viewBox.height).toBe(1500);
+    const viewBoxAndDimensions = MetadataUtils.getViewBoxAndDimensions(nodes, textNodes, new SvgParameters(undefined));
+    expect(viewBoxAndDimensions.viewbox.x).toBe(-700);
+    expect(viewBoxAndDimensions.viewbox.y).toBe(-740);
+    expect(viewBoxAndDimensions.viewbox.width).toBe(1700);
+    expect(viewBoxAndDimensions.viewbox.height).toBe(1500);
+    expect(viewBoxAndDimensions.width).toBe(340);
+    expect(viewBoxAndDimensions.height).toBe(300);
```
NathanDissoubray, line 390: what difference is there between the width and the width of the viewbox ? is the width fixed but the viewbox changeable ?

demo/java/network-viewer-demo-data/src/main/java/com/powsybl/viewer/demo/NadDemoFiles.java
massimo-ferraro: When this diagram was added to the demo, the basic force layout was used for creating it. Here the goal, for testing the PR changes, was to add the width and height attributes to the diagram SVG included in the demo files and the test data. It was needed a small change in the NAD paramenters, but there was no need to change the layout used for creating it.

packages/network-viewer-core/src/network-area-diagram-viewer/metadata-utils.ts
massimo-ferraro: Well, yes, this function is exported, and the result of the function has been changed. So, yes, if the function is used outside the viewer, this can be a breaking change

packages/network-viewer-core/src/network-area-diagram-viewer/metadata-utils.test.ts
massimo-ferraro: Width and height of the SVG (the added parameters) define the size of the viewport, i.e. the window or display area for the SVG. If this attributes are not defined in the SVG provided as input to the NAD viewer, the viewer does compute them, using the SVG content and other input parameters.
The viewbox defines the internal coordinate system, and the part of the SVG you can see in the viewport. Changing the zoom in the NAD viewer (using the viewer zoom functionalities) does change the viewbox, not the viewport.

packages/network-viewer-core/src/network-area-diagram-viewer/metadata-utils.ts
NathanDissoubray: Would you prefer reporting this in the breaking change section of the PR for the https://github.com/powsybl/powsybl-network-viewer/tree/integration/nad_create_svg branch ? It might be lost if we write it here only

massimo-ferraro: I added the information about the change in the utils function in the PR description. I didn't marked the PR as a breaking change, I do not usually mark changes of utils functions, not so rare, as breaking changes

NathanDissoubray: It's a public function, whose name and return type changed. It should be marked as a breaking change.

massimo-ferraro: Well, that's a utility function, used by the viewer and other project classes, like the svg writer. It' not a public method of the viewer, so it should not impact who is using directly the viewer. Of course it is public (exported), otherwise it would not be usable by the viewer. Anyway, I marked the PR as introducing breaking changes, I will do that

NathanDissoubray (approved):

PR #380, opened 2026-04-16
Add unit and browser tests for network map layers, upgrade deck.gl to 9.3

amarant:
- Implemented tests for ArrowLayer, ForkLineLayer, ParallelPathLayer, ScatterplotLayerExt, and LineLayer.
- Added screenshot comparison functionality for visual regression testing.
- Created utility functions for rendering layers and setting up test environments.
- Introduced network and geo data creation functions for testing purposes.
- Added browser setup entrypoint for package-level shared initialization.
- Updated App.tsx to utilize large network and geo data for testing.
- Added new test cases for arrow, fork, parallel path, and scatterplot layers to handle large network scenarios.
- Introduced large network and geo data creation functions in network-fixtures.ts.
- Adjusted browser rendering setup to accommodate larger viewport sizes.
- Enhanced logging for current view state in network-map.tsx.
- Updated Vitest configuration for increased viewport dimensions in tests.
- Updated @deck.gl and @luma.gl dependencies to 9.3.0 versions in package.json and network-map-layers/package.json.
- Add Playwright installation step to CI build process
- Add upload step for browser failure screenshots in CI workflows
- Add image approximate comparison in tests
- Add sonar ignore for test fixtures
- Refactor LineLayer deprecated coordinate system enum to use string literals

**Please check if the PR fulfills these requirements**

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)
- [ ] A PR or issue has been opened in all impacted repositories (if any)

**What kind of change does this PR introduce?**

Add tests for map layers

**What is the current behavior?**

**What is the new behavior (if this is a feature change)?**

**Does this PR introduce a breaking change or deprecate an API?**
- [ ] Yes
- [x] No

PR #431, opened 2026-07-17
Quality fixes on network-map-viewer

### LINKED ISSUE #64: Various quality fixes on network-map-viewer
Comments posted by @capyq in powsybl/powsybl-incubator#267 which were on the code extracted from https://github.com/gridsuite/gridstudy-app
This should be cut later in pieces to separate the easy parts from the most difficult ones (like typescript).

## _Originally posted in https://github.com/powsybl/powsybl-incubator/issues/267#issuecomment-1812200503_
Globaly, it would be better to switch to typescript it will reveal a lot of bug. 
I notice to much confusion between "null" and "undefined".
it's not a javaScript way to handle Object by using the class Map. it's better to use JSON Object.

            
## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393894088_
```js
let labelDistanceInSegment = remainingDistance;
if (goodSegment.idx === 0) {
    labelDistanceInSegment = -alreadyDoneDistance;
} 
```

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393903663_
```js
return equipments ?? [];
```

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393905321_
```js
if (!equipementsToIndex?.length) {
```

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393914702_
```js
if (substationAdded || voltageLevelAdded) {
```
it's more readable

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393919271_
why doing this at the end ? 
this is cost expensive and if it's needed, you should add an else statement on the fullrender check (line 108)

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393936295_
this should be a "const" you never asign it after this line

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393939164_
does "null" and undefined are separate ? if not, you better use this version 
```js
if (!props.network && !props.geoData) {
```

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393955251_
does null and undefined are two expected value ? if not, prefer short version 
```js
    !props.network && !props.geoData && !props.filteredNominalVoltages
) {
```

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393958192_
separate in two lines we don't understand what should not be undefined

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393965658_
this method is never used

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393974367_
you are mixing null and undefined 
1 -  if infos is undefined you also should return null (or undefined)  => if (!infos){...}
2 - why testing infos?.id but not infos.name ?

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393976765_
those are not else if but only if  and you don't need else case because you return each statement

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1393980493_
you shouldn't use tabWith : 4 with the linewith 80. 
so many line goes to 2 lines or more

## _Originally posted in https://github.com/powsybl/powsybl-incubator/pull/267#discussion_r1418502677_
It's not a good practice to use variable to store component React doesn't tag this as react component and re-render it more than needed.

samirromdhani:

Fixes #64

- [ ] Yes
- [x] No

**Other information**:


packages/network-map-layers/src/network/map-equipments.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+        test('removeEquipment of type VoltageLevel should remove it from the VoltageLevels list', () => {
+            // Given
+            const voltageLevelToRemove = 'v1';
+            const removeBranchesOfVoltageLevelMocked = vi.spyOn(equipment, 'removeBranchesOfVoltageLevel');
+            // When
+            equipment.removeEquipment(EQUIPMENT_TYPES.VOLTAGE_LEVEL, voltageLevelToRemove);
+            // Then
+            expect(equipment.getVoltageLevel(voltageLevelToRemove)).toBeUndefined();
+            expect(equipment.getVoltageLevels().map((value) => value.id)).not.toContain([voltageLevelToRemove]);
```
CBiasuzzi: shouldn't it be 
`expect(equipment.getVoltageLevels().map((value) => value.id)).not.toContain(voltageLevelToRemove);`
?
With this fix the test fails, exposing an error with the removeEquipment implementation

samirromdhani: Thanks @CBiasuzzi , good catch!
`removeEquipment` seems needs refactoring. When we call removeEquipment of type VoltageLevel, the voltage level should be removed from the network model (MapEquipments), not just detached from its parent substation. (same logic as Line)
I've updated the implementation to match this specification:
```js
expect(equipment.getVoltageLevels().map((value) => value.id)).not.toContain(voltageLevelToRemove);
expect(equipment.voltageLevelsById.has(voltageLevelToRemove)).toBe(false);
```

packages/network-map-layers/src/network/map-equipments.test.ts
```diff
@@ -0,0 +1,157 @@
...
+            // When
+            equipment.removeEquipment(EQUIPMENT_TYPES.VOLTAGE_LEVEL, voltageLevelToRemove);
+            // Then
+            expect(equipment.getVoltageLevel(voltageLevelToRemove)).toBeUndefined();
+            expect(equipment.getVoltageLevels().map((value) => value.id)).not.toContain(voltageLevelToRemove);
+            expect(equipment.voltageLevelsById.has(voltageLevelToRemove)).toBe(false);
+            expect(removeBranchesOfVoltageLevelMocked).toHaveBeenCalledTimes(1);
+            expect(removeBranchesOfVoltageLevelMocked).toHaveBeenCalledWith(equipment.getLines(), voltageLevelToRemove);
+            expect(equipment.getNominalVoltages()).toStrictEqual([400, 90]);
+        });
```
CBiasuzzi, line 123: still regarding the removeEquipment method:  It seems that the substation's voltageLevels is not updated anymore, after a voltage level is removed. Is this intended?  
If you append here this assertion, it fails apparently: `expect(equipment.getSubstation('s0')?.voltageLevels).toHaveLength(1);`

samirromdhani: Nothing intended, this pr tries to fix some quality issues based on the tests.
I think the expected behavior is to remove the voltage level from both the network model and it's substation, what do you think ?

samirromdhani: Added a fix

packages/network-map-layers/src/network/map-equipments.test.ts
```diff
@@ -0,0 +1,157 @@
...
+            const voltageLevelToRemove = 'v1';
+            const removeBranchesOfVoltageLevelMocked = vi.spyOn(equipment, 'removeBranchesOfVoltageLevel');
+            // When
+            equipment.removeEquipment(EQUIPMENT_TYPES.VOLTAGE_LEVEL, voltageLevelToRemove);
+            // Then
+            expect(equipment.getVoltageLevel(voltageLevelToRemove)).toBeUndefined();
+            expect(equipment.getVoltageLevels().map((value) => value.id)).not.toContain(voltageLevelToRemove);
+            expect(equipment.voltageLevelsById.has(voltageLevelToRemove)).toBe(false);
+            expect(removeBranchesOfVoltageLevelMocked).toHaveBeenCalledTimes(1);
+            expect(removeBranchesOfVoltageLevelMocked).toHaveBeenCalledWith(equipment.getLines(), voltageLevelToRemove);
```
CBiasuzzi: After the vl1 removal, there is a difference between getLine and getLines, apparently.
E.g., with these statements added to this  test:

```
expect(equipment.getLine('l1')).toBeUndefined(); // l1 removed,  l0 remains -> OK
expect(equipment.getLines().map((l) => l.id)).not.toContain('l1'); //  this one fails: both l0 and l1 are still there
```

samirromdhani: You're right, i think `removeBranchesOfVoltageLevel` needs refactoring as well. It should be changed into a void method that updates the whole list of lines (getLines), not just (getLine).

amarant:
following #380 we use the [new deck.gl vitest harness to test the layers](https://deck.gl/docs/api-reference/test-utils/test-layer#example-vitest), it might be useful to put those new tests inside it if needed.

samirromdhani:
> following #380 we use the [new deck.gl vitest harness to test the layers](https://deck.gl/docs/api-reference/test-utils/test-layer#example-vitest), it might be useful to put those new tests inside it if needed.

Done, added the layer lifecycle tests in `LineLayer` in order to fix conflicts. Nice to try testLayer!

packages/network-map-layers/src/network/line-layer.test.ts
```diff
@@ -6,55 +6,106 @@
  */
 
 import { testLayer } from '@deck.gl/test-utils/vitest';
-import { expect, test } from 'vitest';
+import { describe, beforeEach, test, expect, vi } from 'vitest';
 import { createGeoData, createLineForLineLayer, createNetwork } from '../testUtils/network-fixtures.test.utils';
 import { LineFlowMode, LineLayer } from './line-layer';
+import { MapEquipments } from './map-equipments';
+import { MapLineWithType } from '../equipment-types.ts';
```
CBiasuzzi: wouldn't it be better to remove the .ts suffix, from these two imports?

packages/network-map-layers/src/network/map-equipments.ts
```diff
@@ -183,8 +183,7 @@ export class MapEquipments {
             (l) => l.voltageLevelId1 !== voltageLevelId && l.voltageLevelId2 !== voltageLevelId
         );
         branchesList.filter((l) => !remainingLines.includes(l)).forEach((l) => this.linesById.delete(l.id));
-
-        return remainingLines;
+        this.lines = remainingLines;
```
CBiasuzzi: minor issue:  removeBranchesOfVoltageLevel is (probably) never called directly, but this update changes its (implicit) return signature to void. What if we keep  the original
`return remainingLines;  `
and assign that returned value in method removeEquipment, instead?
`this.lines = this.removeBranchesOfVoltageLevel(this.getLines(), equipmentId);`

samirromdhani: Good point, but the method name sounds void-only. What about renaming it to `getBranchesDetachedFromVoltageLevel`?
```javascript
this.lines = this.getBranchesDetachedFromVoltageLevel(this.getLines(), equipmentId);

```

packages/network-map-layers/src/network/line-layer.test.ts
samirromdhani: Done

packages/network-map-layers/src/network/map-equipments.ts
CBiasuzzi: it could be; but it has be documented as a change in this case....  as said, who knows if anyone (in the world)  is calling that method directly (probalby no-one)

samirromdhani: OK, as it's out of scope for this issue, I'll add a TODO to make it private and rename it in a future change

amarant (approved):

sBouzols:
Do we really need to add `@testing-library/react` @samirromdhani ? 
Otherwise everything seems ok to me.

samirromdhani:
> Do we really need to add `@testing-library/react` @samirromdhani ? Otherwise everything seems ok to me.

Hi @sBouzols, 
Yes it is needed by the new `useNameOrId` tests. Since `useNameOrId` calls `useCallback`, it is a react hook and cannot be called directly outside a react render context.

PR #436, opened 2026-07-21
[nad-viewer] create style and footer for legends generated from metadata

### LINKED ISSUE #435: [nad-viewer] Voltage level legends not fully created from metadata (style, footer)

When the voltage level legend is created from metadata:

- if a style attribute is defined (e.g. when created by diagram's custom style provider) the colored squares do not get the right style from the metadata.
- if a legendFooter is defined in metadata, it is not added to the legend box.

In the generated voltage level legend box

- the bus legends square style should match the style defined in metadata, if available.
- the legendFooter, if available, should be displayed (below the bus legend section).

### Describe the steps

This issue can be reproduced in the demo's "Custom style provider" diagram

- Edit demo's nad.ts and enable the adaptiveTextZoom for the "Custom style provider" diagram (e.g., in NadSvgPstHvdcCustomExample section, add the parameter adaptiveTextZoom: { enabled: true, threshold: 1500 } )

- Add a legendFooter to  the "Custom style provider" demo JSON: open NadDemoFiles.java and edit drawFourSubstationsCustomLabelAndStyle method to add a legend footer, e.g. "S1VL1 footer1"
```
	  var vl1Legend = new VoltageLevelLegend(
		List.of("S1VL1 description1"),
		List.of("S1VL1 footer1"),
		Map.of("S1VL1_0", "S1VL1_0 description"));
```		

- Regenerate the demo data
- Start the demo
- Zoom-out the "Custom style provider " NAD until the legends disappear, then zoom-in to trigger the adaptive zoom, and examine the created legend boxes.  The legends squares color is 'black'. The legendFooter is not in its legend box , anymore.

### Environment

_No response_

### Relevant Log Output

_No response_

### Extra Information

_No response_

CBiasuzzi:

Fixes #435 

- [ ] Yes
- [X] No

**Other information**:

### DIFF
--- a/packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.test.ts
+++ b/packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.test.ts
@@ -8,12 +8,12 @@
 
 import { NadViewerParametersOptions } from './nad-viewer-parameters';
 import { NetworkAreaDiagramViewer } from './network-area-diagram-viewer';
+import { BusNodeMetadata, NodeMetadata, TextNodeMetadata } from './diagram-metadata';
 
 describe('Test network-area-diagram-viewer', () => {
+    const container: HTMLDivElement = document.createElement('div');
     // SVG aren't loaded properly in DOM with Vitest. Has to be enriched...
     test('nad creation', () => {
-        const container: HTMLDivElement = document.createElement('div');
-
         const nadViewerParametersOptions: NadViewerParametersOptions = {
             minWidth: 0,
             minHeight: 0,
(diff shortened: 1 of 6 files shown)


packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.test.ts
```diff
@@ -40,4 +40,60 @@ describe('Test network-area-diagram-viewer', () => {
...
+    test('legend boxes: bus nodes style attributes should be used, when defined in metadata', () => {
+        const nadViewer: NetworkAreaDiagramViewer = new NetworkAreaDiagramViewer(container, '', null, null);
+        const node: NodeMetadata = {
+            svgId: '0',
+            equipmentId: 'EQUIPMENT_ID_0',
+            x: 0,
+            y: 0,
+        };
+        const busNodes: BusNodeMetadata[] = [busNode(0, '400 kV', 'background:red; fill:red; stroke:black;')];
+        const legendBox = nadViewer['createLegendBox'](textNode, busNodes, node);
```
NathanDissoubray, line 63: createLegendBox is private. Do we need a public constructor, or is it usual in typescript to test private methods like that ?

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
```diff
@@ -1879,12 +1879,18 @@ export class NetworkAreaDiagramViewer {
...
+                newBusLegendElement.setAttribute('style', busNode.style);
+            }
             const textNode = document.createTextNode(busNode.legend ?? '');
             newBusDivElement?.appendChild(newBusLegendElement);
             newBusDivElement?.appendChild(textNode);
             newDivElement.appendChild(newBusDivElement);
         }
+        node.legendFooter?.forEach((footer) => {
+            newDivElement.appendChild(this.createTextHeader(footer));
+        });
```
NathanDissoubray, line 1892: this is the same thing as for legendHeader a bit above, line 1866
Could we make an helper function for that maybe ? If it's only in two places, maybe not needed, but if it's in other places it would be good.
Also, I think you should put this part just after the legendHeader (and not after the loop on busNodes)

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.test.ts
CBiasuzzi: It seems it is an acceptable way to test a private method (note also that the very same approach is already used by PR  430 )

packages/network-viewer-core/src/network-area-diagram-viewer/network-area-diagram-viewer.ts
CBiasuzzi: Thanks for the feedback.

It’s only used in those two places,  I think it’s probably not worth extracting a helper for just those two lines that are involved.

Why shouldn’t I put the footer after the loop on the nodes? A footer belongs at the bottom, don't you think?
(see also this example: https://github.com/powsybl/powsybl-diagram/blob/main/network-area-diagram/src/test/resources/custom_label_provider.svg )

NathanDissoubray: I was thinking of putting everything "header" related (whether that is top or bottom) at the same place in the code, but I guess the things are drawn in the order in which they are written in the code, in which case it's better to put it after the loop indeed

CBiasuzzi: Yes, here the elements are rendered in the order they're processed in the code.  It basically mimics what happens in powsybl-diagram's SvgWriter.

NathanDissoubray (changes requested):
I don't see any test file showing the colored squares that were mentionned in #435 
Would it be possible to have an example ?

CBiasuzzi:
I've added to the issue #435 a couple of screenshots illustrating the issue and the expected behavior, as mentioned in the issue description.

Note that in this fix I updated the demo json to include a footer attribute (there were no footers in the original demo file). The style attributes for  the bus nodes (declaring the custom color for the squares) however were already in the metadata - no need to add the styles. I also activated the adaptive zoom feature (in the main branch it is not active).
I hope this clarifies it a little bit more.

NathanDissoubray (approved):

PR #443, opened 2026-07-30
Select a voltage level by clicking its label in a NAD

flomillot:
## Context

In a network area diagram, only the voltage level node itself reacted to a left click.

`isSelectable` only accepted children of `g.nad-vl-nodes`. Text nodes (children of `g.nad-text-nodes`) were hoverable and draggable but never selectable, so a left click on a label never reached `onSelectNodeCallback`.

## Changes

- `isSelectable` now also accepts text nodes, reusing the existing `isTextNode` predicate.
- `callSelectNodeCallback` resolves a selected text node to the voltage level node it labels, through `textNodes[].vlNode`, so the callback keeps reporting the node's own `equipmentId` and `svgId`.


CBiasuzzi (commented):
With this change, a consumer for the hovering callback will not be able to distinguish anymore between the hovering on a VL node and the hovering on the VL text node; 
I suppose this is intended

> "A text node is not an equipment type ..."


However, the right click callback was not touched, so a consumer for that callback will still get

- equipmentType: VOLTAGE_LEVEL   (right click on a VL node)
- equipmentType: TEXT_NODE  (right click on a VL text node)

is this intended, as well?

flomillot:
Hello you're right, I haven't search enough and can do it in the consumer directly which is much cleaner, I changed it. 
Is the resting part OK ?
Thank you.

CBiasuzzi (approved):
It looks good to me, with respect to this PR’s target

massimo-ferraro (approved):

ayolab (approved):

